PR #12, opened 2019-05-08
references and mixed usage from JavaSE

struberg:


struberg:
Hi folks! The eclipse eca check has a known bug. I'm registered at EF with my apache address. You can verify it with the eclipse tools that I have an active ECA on file. 
Probably the script blurps up because I am in volved in multiple orgs and have registered multiple addresses. Seems to be a pretty random issue afaict.

@bshannon 
That means that `javax.transaction.xa.*` are JavaSE and will not be stripped from the JDK in the future?
And they are _not_ to be included in the jakarta-transaction spec?

Regarding `javax.security` it's pretty much the same. We need to clearly define which subpackages are in jakarta and which are part of SE security.

bshannon:
Yes, javax.transaction.xa is SE only, not in the Jakarta Transactions spec.

Look at the Java SE and Java EE platform javadocs to see which javax.security subpackages are in each.

keilw:
And what makes it even more sophisticated are name changes of certain specs like "Java EE Security" simply into "Jakarta Security" where the added package `javax.security.enterprise` is most likely to become just `jakarta.security`, so it is not a simple replacement of "javax" to "jakarta" in several situations.

struberg:
This is exactly something I would void. Let's have clean rules. Once we start with anything else than s/javax/jakarta/ then it will end up as big pain for users. I also see no technical reason for it tbh.

ggam:
+1

ivargrimstad:
> Seems that I've edited the file with my yahoo address and not with my apache.org which I'm registered with at the EF. Anyway, you can verify my address struberg @ apache.org with the eclipse tools and see that I have a valid ECA on file

@struberg Yes, the apache mail is valid. Not sure how I can bypass the automatic check here though... Maybe @eclipsewebmaster or @waynebeaton have some tips?

struberg:
I gonna ship a new PR with my apache.org email. Just at customer right now, so no infra avail.

PR #15, opened 2019-05-14
Initial binary compatibility levels outline

starksm64:
This is an initial attempt at a binary levels of compatibility document that tries to outline the complexity of each level

### DIFF
--- a/namespace/bcl.adoc
+++ b/namespace/bcl.adoc
@@ -1,0 +1,46 @@
+= Binary Compatibility Levels
+
+This is a classification of the levels of binary compatibility support for Java EE 8 classic applications in terms of the level of difficulty/complexity of the associated implementation and tools.
+
+== Level 0 - Jakarta EE 8 Only
+* No transformation required
+* Requires a Jakarta EE 8 container
+* No mandated support for backwards compatibility
+** Up to implementors to decide what level of support to provide
+
+== Level 1 - Jakarta EE 8 Translation to EE 9
+* Requires a direct translation of javax.* packages to jakarta.* analog
+* Requires transformation of META-INF/services to jakarta analogs
+* Requires transformation of properties files referencing javax based properties
+* Requires transformation of server specific properties that are classic javax based properties
+** This could be a specfic build or deployment step that processes an EE 8 app
+* Requires only a Jakarta EE 9 container
+* Requires the big bang javax to jakarta approach
+** This simplifies the EE9 container
+
+== Level 2 - Colocated Jakarta EE 8/9 with Isolation
+* Requires an isolated Jakarta EE 8 container
+* Requires an isolated  Jakarta EE 9 container
+* No interaction of EE 8 and EE 9 components
+* Requires the big bang javax to jakarta approach
+** Simplifies the EE9 container implementation in terms of isolating javax packages from jakarta packages.
+
+== Level 3 - Jakarta EE 9 With Mapping from EE 8
+* Requires a Jakarta EE 9 container
+* Requires a Jakarta EE 8 to EE 9 compatibility layer
+* Requires either compatibility or mapping layer for services and properties
+* Requires mapping of server specific properties that are classic javax based properties
+* Components interact as peers as EE8 components are exposed as EE9
+** Unclear what holes exist in a compatibility layer solution
+*** Likely some runtime overhead cost
+* Could work with either big bang or incremental approach
+** Incremental approach has increasing complexity as more javax APIs evolve
+*** How to track how an EE8 app is handled beyond EE 9 is the issue
+** Big bang would have the simplest evolution path beyond EE 9
+
+== Level 4 - Colocated Jakarta EE 8 and EE 9 with API Mixing
+* Supports side by side javax.* and jakarta.* components in one application
+* Requires a Jakarta EE 9 container with native support for EE 8 components
+* Theoretically does not require any pre-processing of EE 8 application
+* Allows incremental approach
+* Likely large technical challenges in bridging between the EE 8 and 9 components


namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+** This could be a specfic build or deployment step that processes an EE 8 app
+* Requires only a Jakarta EE 9 container
+* Requires the big bang javax to jakarta approach
+** This simplifies the EE9 container
+
+== Level 2 - Colocated Jakarta EE 8/9 with Isolation
+* Requires an isolated Jakarta EE 8 container
+* Requires an isolated  Jakarta EE 9 container
+* No interaction of EE 8 and EE 9 components
+* Requires the big bang javax to jakarta approach
```
gregw: not necessarily.   The isolated EE9 can be whatever incremental or big bang EE spec it wants to be, because it is isolated.

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+** Up to implementors to decide what level of support to provide
+
+== Level 1 - Jakarta EE 8 Translation to EE 9
+* Requires a direct translation of javax.* packages to jakarta.* analog
+* Requires transformation of META-INF/services to jakarta analogs
+* Requires transformation of properties files referencing javax based properties
+* Requires transformation of server specific properties that are classic javax based properties
+** This could be a specfic build or deployment step that processes an EE 8 app
+* Requires only a Jakarta EE 9 container
+* Requires the big bang javax to jakarta approach
```
gregw: not necessarily. The translation tools could evolve with any incremental approach

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+This is a classification of the levels of binary compatibility support for Java EE 8 classic applications in terms of the level of difficulty/complexity of the associated implementation and tools.
+
+== Level 0 - Jakarta EE 8 Only
+* No transformation required
+* Requires a Jakarta EE 8 container
+* No mandated support for backwards compatibility
+** Up to implementors to decide what level of support to provide
+
+== Level 1 - Jakarta EE 8 Translation to EE 9
```
gregw: a niggle about the ordering.  I see this as coming after your `Jakarta EE 9 With Mapping from EE 8` as a level as it belong to the variations were we no longer have an EE8 container.

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+* Requires a direct translation of javax.* packages to jakarta.* analog
+* Requires transformation of META-INF/services to jakarta analogs
+* Requires transformation of properties files referencing javax based properties
+* Requires transformation of server specific properties that are classic javax based properties
+** This could be a specfic build or deployment step that processes an EE 8 app
```
gregw: A build step would produce an EE9 binary that could live a long time... but there would be no source code and no migration path to EE9.1 or beyond for that binary.

A deployment step would have a startup time cost and the translation tools would need to be maintained for EE9.1 and beyond, which could become increasingly difficult.    If it is a deployment step, it should perhaps be deprecated from the start and produce a warning that the application should be updated.

Another option is a source generation tool to generate EE9 sourcecode from EE8 source code... but that is perhaps beyond our scope?

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+** Incremental approach has increasing complexity as more javax APIs evolve
+*** How to track how an EE8 app is handled beyond EE 9 is the issue
+** Big bang would have the simplest evolution path beyond EE 9
+
+== Level 4 - Colocated Jakarta EE 8 and EE 9 with API Mixing
+* Supports side by side javax.* and jakarta.* components in one application
+* Requires a Jakarta EE 9 container with native support for EE 8 components
+* Theoretically does not require any pre-processing of EE 8 application
+* Allows incremental approach
+* Likely large technical challenges in bridging between the EE 8 and 9 components
```
gregw: the pixies and unicorns will help!

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
+= Binary Compatibility Levels
+
+This is a classification of the levels of binary compatibility support for Java EE 8 classic applications in terms of the level of difficulty/complexity of the associated implementation and tools.
+
+== Level 0 - Jakarta EE 8 Only
```
bshannon: I don't understand this level.  Is this a product that supports only Jakarta EE 8?
Obviously the binary compatibility with Jakarta EE 8 for such a product would
be "complete".  There would be no issue of backwards compatibility.  There
would be no choice for implementers to decide.

Obviously you must mean something else by this level but I don't understand
what it is.

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+* No transformation required
+* Requires a Jakarta EE 8 container
+* No mandated support for backwards compatibility
+** Up to implementors to decide what level of support to provide
+
+== Level 1 - Jakarta EE 8 Translation to EE 9
+* Requires a direct translation of javax.* packages to jakarta.* analog
```
bshannon: I assume what you have in mind here is effectively making all the javax.* classes be
"aliases" for the corresponding jakarta.* classes.  Two names, one class.

namespace/bcl.adoc
bshannon: Doing this at deployment time has the advantage of not requiring any tools to change
and paying the cost only once for the lifetime of the deployment.

Doing this at build time or run time has the advantage that it could be used by applications
outside of an application server, e.g., using "standalone" versions of the new jakarta.* APIs.

It's not clear to me why you think maintaining the translation tools will become increasingly
difficult.

If binary compatibility is defined by a spec, that spec might have limited lifetime, but I
would expect that vendors will need to maintain any translation tools for a very long time.

While source level tools might be outside our scope, we should definitely work with tool vendors to
create the refactoring tools that would allow this to be done at the source code level.

namespace/bcl.adoc
bshannon: I agree, I don't see why this requires a big bang.

namespace/bcl.adoc
bshannon: I agree.  The isolated jakarta.* runtime could grow over time.

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+** This simplifies the EE9 container
+
+== Level 2 - Colocated Jakarta EE 8/9 with Isolation
+* Requires an isolated Jakarta EE 8 container
+* Requires an isolated  Jakarta EE 9 container
+* No interaction of EE 8 and EE 9 components
+* Requires the big bang javax to jakarta approach
+** Simplifies the EE9 container implementation in terms of isolating javax packages from jakarta packages.
+
+== Level 3 - Jakarta EE 9 With Mapping from EE 8
```
bshannon: I don't understand how this is different from Level 1.

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+* Requires a Jakarta EE 9 container
+* Requires a Jakarta EE 8 to EE 9 compatibility layer
+* Requires either compatibility or mapping layer for services and properties
+* Requires mapping of server specific properties that are classic javax based properties
```
bshannon: Are you imagining that servers are defining properties in the javax.* namespace
and that those properties will need to be mapped along with the spec-defined
javax.* properties?

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+* Components interact as peers as EE8 components are exposed as EE9
+** Unclear what holes exist in a compatibility layer solution
+*** Likely some runtime overhead cost
+* Could work with either big bang or incremental approach
+** Incremental approach has increasing complexity as more javax APIs evolve
+*** How to track how an EE8 app is handled beyond EE 9 is the issue
```
bshannon: I don't understand this.

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+** Big bang would have the simplest evolution path beyond EE 9
```
bshannon: I don't understand this.

namespace/bcl.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+== Level 4 - Colocated Jakarta EE 8 and EE 9 with API Mixing
+* Supports side by side javax.* and jakarta.* components in one application
```
bshannon: In this scenario the javax.* class and the corresponding jakarta.* class are different
classes, with no inheritance relationship, right?

namespace/bcl.adoc
bshannon: I'm not sure what you have in mind by "bridging".

namespace/bcl.adoc
gregw: Bill, both Scott and I identified a mode 0 as just running an EE8 container/server as a way to support EE8 applications.   Yes it is kind of obvious to us, but I think it is important for us to say as much to the community - ie that we will continue supporting our EE8 releases for sometime and that there is no immediate prerogative to upgrade to an EE9 capable container.

bshannon: I think it just confuses the discussion of whether or how to provide compatibility
with Jakarta EE 8 in a Jakarta EE 9 or later product.

gregw: I think it precisely defines how many vendors are going to provide compatibility for Jakarta EE8, once EE9 is released.  We are going to maintain our pure EE8 platforms for a lot longer than a normal version change because there will be some applications that cannot or willnot change (or be changeable) to the new APIs.   With APIs like servlet-api, many version increments have been very minor and thus there has been very little impediment to the end-of-life of previous versions as binary compatibility was trivial.  This will not be the case and vendors are going to have to decide which of the other modes - if any - they will implement.   Mode 0, will be the primary mode for many, perhaps the only mode.

namespace/bcl.adoc
starksm64: I was thinking that isolation of the Jakarta EE 9 container could be more problematic wtihout the big bang, but I guess it could be managed.

namespace/bcl.adoc
starksm64: Here I do see the complexity of the translation increasing if there is a mixture of javax/jakarta packages vs a straight translation to jakarta. Could say big bang is level 1 and incremental level 1.5.

namespace/bcl.adoc
starksm64: I see. My ordering was more about perceived complexity to implement vs a how much Java EE 8 existed at runtime.

namespace/bcl.adoc
bshannon: That's not providing compatibility.  That's like saying Java is compatible with C# because I can tell you to go use that C# product to run your C# application.

namespace/bcl.adoc
bshannon: I don't see why that increases the complexity.  The *size* of the translation rules will increase as you add more javax APIs that have to be translated to jakarta APIs, but in general that wouldn't increase the complexity unless the additional javax APIs introduced previously unseen compatibility challenges.  Certainly having *fewer* javax APIs that needed to be translated would not *increase* the complexity.

namespace/bcl.adoc
gregw: To run with your analogy, mode zero is saying to C# vendors that they may well be moving to become Go runtime vendors, but they have to keep maintaining their C# runtimes until such time is that fancy new Go runtime can also seamlessly run C# applications.

bshannon: We're not telling vendors what they *have* to do.  At this point, we're *not* expecting
that compatibility is *required*.

So yes, the C# runtime vendor can abandon all their C# customers if they think that's
a better business model, or they can tell them that there's no path for them to take
advantage of new features or performance in the vendor's new Go runtime product so
they should keep using the C# product, or they can build a Go runtime product that provides
some compatibility for their C# applications and allows the customers to benefit from the
new features and performance in the vendor's Go runtime product.  It's the vendor's choice.

gregw: Exactly, there are lots of options and this document is about enumerating them.    So perhaps there is also a mode -1,  so using my documents numbering:
 + mode -1 - no EE8 support, EE9 only
 + mode 0 - separate EE8 and EE9 servers
 + mode 1 - isolated EE8 and EE9 containers in the same server
 + mode 2 - EE8 and EE9 containers in same server with shared JNDI space
 + mode 3 - EE8 wrapped as EE9 in a EE9 only server
 + mode 4 - EE8 translated to EE9 in EE9 only server
 + mode 5 - magical mixed EE8 and EE9 container.

I see no harm in enumerating all the possibilities, and I see benefit in being clear in calling out that mode 0 is a likely outcome/step for many vendors.  We are not telling vendors which modes they must implement, but in discussing how our EE ecosystem moves forward it is important to have a complete picture.

bshannon: Maybe this is a matter of perspective...

For me, the important issue is what binary compatibility does a Jakarta EE 9 **product**
support.  You seem to be answering the question of what binary compatibility does a
Jakarta EE 9 **vendor** support.  The former is a technical problem.  The latter can be
dealt with as just a business issue.

From a vendor perspective, the choices then become...

- mode -1 - vendor provides only an EE 9 product
- mode 0 - vendor provides separate EE 8 and EE 9 products
- mode 1 - vendor provides a single product that includes isolated EE 8 and EE 9 containers

and so on.  In my mind, mode -1 and mode 0 are "no binary compatibility".
I'm much more interested in the technical choices for how **to** provide binary compatibility.

gregw: Bill, I totally agree that modes -1 and 0 are non technical modes that are "no binary compatibility".    Also agree that it is more important to work out technical details of modes >0.
I just think that from a project point of view it is important to enumerate all the possibilities and "no binary compatibility" is a valid possibility.     Anyway... I think we've established we have different points of view on this one... let others adjudicate and let's focus on the technical details of modes >0

bshannon:
@starksm64 Can you resolve the issues in this conversation so we can merge this?
Or close the PR if you're no longer interested.

Thanks.

bshannon:
@keilw I don't know why you merged this PR that was still under discussion with
many open issue but please un-merge it immediately.  Thanks.

PR #20, opened 2019-05-22
add all specs todo

rieonke:
Signed-off-by: Rieon Ke <rieon@rieon.cn>


rieonke:
Add all specs todo placeholder, so we can work parallelly.

PR #23, opened 2019-05-29
Clarify that only javax needs to change in package name

kwsutter:
I think we need to be very clear that only the "javax" needs to change in the package name.  The other names and acronyms used in the package names do not have to change.

### DIFF
--- a/namespace/README.adoc
+++ b/namespace/README.adoc
@@ -33,7 +33,8 @@ Open discussion is currently taking place on
 Other proposals should incorporate the following considerations and goals:
 
 * The new namespace will be `jakarta.*`
-* APIs moved to the `jakarta` namespace maintain class names and method signatures compatible with equivalent class names and method signatures in the `javax`.* namespace.
+* APIs moved to the `jakarta.\*` namespace maintain package names (except for the `javax`), class names, and method signatures compatible with equivalent package names, class names, and method signatures in the `javax.*` namespace.
+For example, `javax.jms.JMSConsumer.getMessageListener()` would move to `jakarta.jms.JMSConsumer.getMessageListener()`.
 * Even a small maintenance change to an API would require a `javax` to `jakarta` change of that entire specification. Examples include:
 ** Adding a value to an enum
 ** Overriding/adding a method signature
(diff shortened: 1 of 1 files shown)


kwsutter:
> I would phrase it differently, e.g. that other parts "may change", especially the `javax.jms` example is a clear case, where the project team together with others involved like the PMC, etc. agreed, that renaming the project to "Jakarta Messaging" would also advise to change the second level of the package name to something like "jakarta.messaging" or "jakarta.message" instead of "jakarta.jms".

Werner, note that I am not proposing any changes to the current wording other than emphasizing the Package rename of javax to jakarta.  Any other changes to the Package, Class, or Method names should be considered very carefully.  And, continues to be discussed on the PMC mailing list...

Also, I can't seem to figure out what specific Changes you requested.  I don't see any comments or updates to the PR which I can action.  Thanks.

keilw:
Well especially the sentence 
>* APIs moved to the `jakarta.\*` namespace maintain package names (except for the `javax`), class 
> names, and method signatures compatible with equivalent package names, class names, and method > signatures in the `javax.*` namespace. 

should be more like
>* APIs moved to the `jakarta.\*` namespace may keep package names (except for the `javax`), class 
> names, and method signatures compatible with equivalent package names, class names, and method > signatures in the `javax.*` namespace. 

Because as it stands that sentence suggests there is no flexibility and variations, and there are many cases (Security, JMS or formerly called JAX-RS just three most popular) where the name change won't be as linear. If you can think of a better example than the `javax.jms` class that would also be good, else it would confuse people.

bshannon:
I agree with Werner.

kwsutter:
JMS is actually a perfect example of why we should not change anything other than the `javax` to `jakarta`.  Following Werner's example, the ripple effect is not good...

`javax.jms.JMSProducer.getJMSType()` becomes something like...
`jakarta.messaging.MessageProducer.getMessageType()`  ???

Oh, but wait, here is already an existing `javax.jms.MessageProducer` class, so this just gets ugly fast.

keilw:
Actually doing it consequent it would have to be 
`jakarta.messaging.MessagingProducer.getMessagingType()` ;-)

Nobody said, it would be easy, but renaming only the Maven artifactId or Java package and then leave the "JMS" brand in all the other elements would be inconsequent. JMS is probably the most extreme case with 20 elements of the API starting with JMS*. There is also EJB with about half a dozen and JCA, those look like the only cases, check it out yourself: https://javaee.github.io/javaee-spec/javadocs/

PR #24, opened 2019-06-04
Adding an option 3

rmonson:


rmonson:
It's not orthogonal.  It's a refinement of the BigBan.

The BigBang says some, but not all of the javax packages can be switched over. I'm saying that they all must be (except those in Java SE) and that binary compatibility profile would also be required.

bshannon:
So the only difference here is that you're modifying Big Bang's "some or all" to be "all"?
I suppose we could add a 1.1 option that says "all" and a 1.2 option that says "some",
but that doesn't seem helpful at this point.

I don't see how the binary compatibility proposal is *not* orthogonal to the other options.
Any of them could make it required or not.

bshannon:
I don't think this is still relevant so I'm closing it.  Reopen or resubmit if I got it wrong.

PR #26, opened 2019-06-05
new proposal 4 to consider

kwsutter:
This is another alternative to the incremental approach.  Discussing the incremental vs big bang approaches within IBM produced this alternative.  It was briefly brought up on a recent Spec Committee call and it was decided that we should offer it as an alternate proposal.  Comments welcome!

### DIFF
--- a/namespace/proposal-04-extend.adoc
+++ b/namespace/proposal-04-extend.adoc
@@ -1,4 +1,4 @@
-# Proposal 3: Extend javax in Jakarta EE 9 and beyond
+# Proposal 4: Extend javax in Jakarta EE 9 and beyond
 
 The `javax` APIs would live "forever".
 All of the new work in Jakarta EE would extend or build upon the existing `javax` APIs.
--- a/namespace/proposal-04-extend.adoc
+++ b/namespace/proposal-04-extend.adoc
@@ -1,0 +1,34 @@
+# Proposal 3: Extend javax in Jakarta EE 9 and beyond
+
+The `javax` APIs would live "forever".
+All of the new work in Jakarta EE would extend or build upon the existing `javax` APIs.
+This would be similar to what https://microprofile.io[MicroProfile] has accomplished with their https://github.com/eclipse/microprofile-rest-client[Rest Client] and https://github.com/eclipse/microprofile-context-propagation[Concurrency/Context Propagation] components.
+The Rest Client has provided a feature extension to Java EE JAX-RS, while the Concurrency/Context Propagation project has provided a feature extension to Java EE Concurrency Utilities.
+
+- All existing Java EE APIs would stay in the `javax` namespace.
+- New or modified APIs would be in the `jakarta` namespace.
+- Jakarta EE would be a mix of `javax` and `jakarta` packaged APIs for the foreseeable future.
+- When updates are proposed for an existing `javax` API, the first goal is to provide the updates via a `jakarta` extension.
+- If the proposed updates can not be a proper extension and updates to the `javax` API are required, then that API would transition from `javax` to `jakarta`.
+This should be the exception, not the rule.
+
+Pros:
+
+- Easier for customer acceptance since their existing applications will continue to run.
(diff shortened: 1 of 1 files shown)


namespace/proposal-04-extend.adoc
```diff
@@ -0,0 +1,34 @@
+# Proposal 3: Extend javax in Jakarta EE 9 and beyond
```
bshannon: Proposal 3 or 4?  :-)  We need to resolve whether Richard's proposal is really different enough from the existing proposals.

namespace/proposal-04-extend.adoc
```diff
@@ -0,0 +1,34 @@
...
+- Potential for less impact from API change overall -- minimize the API migration.
+- Easier to provide more functional Jakarta EE releases in an efficient manner
+
+Cons:
+
+- Prolonged coordination, cost and complexity to industry affecting conversion tools, users, enterprises, cloud vendors, IDE creators, platform vendors, trainers and book authors.
+- Use of `javax` namespace prolonged (could also be considered a benefit).
+- Difficulty in knowing whether a given API is in the `javax` or the `jakarta` namespace.
+- Transitive dependencies to other specifications makes this option harder for platform implementors although not for consumers.
+- Restrictions on what Java SE implementation can be used for certification
```
bshannon: More cons:

- Additions in the `jakarta` namespace may not change the behavior of existing `javax` APIs in a way that would be incompatible with the Java EE specs.
- Difficulty of adding new functionality without changing existing API.
- Unexpected and unpredictable move to `jakarta` namespace when existing API needs to change to support new functionality.

namespace/proposal-04-extend.adoc
kwsutter: Yep, already found that one with my second commit...  :-)

namespace/proposal-04-extend.adoc
kwsutter: Bill, it looks like @keilw already merged this PR...  But, let's talk about your comments and I'm willing to update the list of Cons...

- Not sure I understand your first bullet.  Do you have too many negatives?  Should it say "Additions in the `jakarta` namespace may change the behavior of existing `javax` APIs in a way that would be incompatible with the Java EE specs."  (removed the "not")
- The second bullet is true. I'll add that.
-  Is this third bullet different from the "Difficulty in knowing..." Con?

bshannon: Yes, I should've been clear that I approved it but left it unmerged to allow the discussion to continue.

No, the first bullet is correct.

Just like you can't add an `org.apache.System.disableArrayBoundsChecking` API and still be Java SE compatible, you can't add a `jakarta.*` API that would change the behavior of an existing `javax.*` API in a way that would violate the existing spec.

The third bullet was more covering the case where you're using an existing `javax.*` API in
one release and then in the next release it has moved into `jakarta.*`.  A good compatibility
story might make that transparent, but you still have the issue of where to look for the
documentation, etc.

Maybe I'm not clear on whether you're proposing that when an individual class needs to
be updated you would move just that class to `jakarta.*`, or the need to update that class
would require moving all classes in the spec to` jakarta.*`.  The former would be very confusing.
The latter would still be surprising because some updates to a spec would not move classes
and then all of a sudden for no obvious reason the entire API would jump to `jakarta.*`.

kwsutter: @bshannon Since the bullets are listed under the "Con" heading, then the "con" is that any additions to the jakarta.* namespace could change the behavior of an existing javax.* API, which would not be good.  There is nothing wrong with your statement as-is, except it's listed under the "Con" heading -- which then makes it difficult to parse.  I understand what you are getting at, I'll see if I can craft the wording that allows it to make sense as a "con".

My take is that if you have to modify a javax.* class, then you have to move the whole API to jakarta.*.

kwsutter: As I was updating the proposal, I decided to go with your original writing of the Con about extending and yet not violating the Java EE spec.  And, I added your second Con statement about this extension idea being more difficult.  Thanks for your help.

PR #27, opened 2019-06-10
Big Bang: remove the ability to freeze specs & clarify pruning candidates

anthonyvdotbe:
The ability to leave specs frozen in the `javax` namespace is a source of contention (see e.g. [here](https://www.eclipse.org/lists/jakartaee-platform-dev/msg00423.html)). From looking at the list of specs in Java EE 8, I believe the only reasonable candidates for freezing would be the specs that were already optional. But optional specs can equally well be pruned. So I'd rather either migrate or prune them: it eliminates the contention, and removing all specs from the `javax` namespace would bring about additional advantages.

(Note that this would also bring the proposal in line with the proposed option 3 (#24) w.r.t. what to move)

Signed-off-by: Anthony Vanelverdinghe <anthonyv.be@outlook.com>

### DIFF
--- a/namespace/proposal-01-bigbang.adoc
+++ b/namespace/proposal-01-bigbang.adoc
@@ -4,8 +4,7 @@ The heart of this proposal is to do a one-time move of API source from the `java
 
 Were we to take this path, a compelling approach would be to do the namespace rename and immediately release this as Jakarta EE 9. Additional modifications would be put into a Jakarta EE 10 which can be developed in parallel, without further delays.
 
-- Some or all Jakarta EE APIs under `javax` would move immediately into `jakarta` as-is.
-- Any packages not moved from `javax` to `jakarta` could be included in Jakarta EE, but would be forever frozen and never move to the `jakarta` namespace.
+- All Jakarta EE APIs under `javax` would either move immediately into `jakarta` as-is, or be pruned from Jakarta EE.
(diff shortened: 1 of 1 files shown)


anthonyvdotbe:
If I understood @Emily-Jiang correctly [here](https://www.eclipse.org/lists/jakartaee-platform-dev/msg00427.html), she actually sees this decision of "what to move" as a prerequisite for comparisons between Big Bang and Incremental. So that's why I'm proposing these refinements: they clarify that everything will be moved, possibly except for a few pruning candidates (for which, I believe, it'll be easy to reach consensus on whether to migrate or prune).

Personally, I've always thought of Big Bang as "everything", both in terms of "when" and "what", and believe the ability to freeze does more harm than good.

bshannon:
I'll leave it to @dblevins to decide if this is just a clarification that matches what he originally
intended when he wrote up the Big Bang proposal, or if this is a refinement that we need to
discuss further.

dblevins:
This would be a refinement.  A very valid one, but still the discussion isn't over on if we do a one-time change where some APIs stay frozen.

@anthonyvdotbe I've copied this proposal as proposal-01a-bigbang-all.adoc so you can redo your PR on that one.  In the parent bigbang proposal the word "all" in the opening phrase "Some or all" is now linked to proposal-01a-bigbang-all.adoc.  Maybe someone out there can submit a proposal-01b-bigbang-some.adoc and we can have a similar refinement down that path.

anthonyvdotbe:
Thanks @dblevins, I've created #30.

PR #29, opened 2019-06-17
Added webgraphviz data

gregw:
Processed data so it can be graphed.


joakime:
If you add ```rankdir="LR";``` then the result is easier to read.

jakarta

bshannon:
I'm going to merge this now.  @joakime if you want the rankdir change, submit a new PR.

PR #33, opened 2019-06-29
Removed Maven Enforcer JDK8

ggam:
I've successfully compiled the project with JDK 12.

### DIFF
--- a/specification/pom.xml
+++ b/specification/pom.xml
@@ -70,8 +70,8 @@
                         <configuration>
                             <rules>
                                 <requireJavaVersion>
-                                    <version>[1.8.0,1.9.0)</version>
-                                    <message>You need JDK8 or lower</message>
+                                    <version>[1.8.0,)</version>
+                                    <message>You need JDK8 or later</message>
                                 </requireJavaVersion>
                             </rules>
                         </configuration>


pzygielo:
@ggam - but enforcer enforces from two sides, and I think your success with JDK12 allows for following change:
```
<requireJavaVersion>
    <version>[1.8.0,)</version>
```
The lower limit can't be reduced, as with JDK7 `specification` will definitely not build due to dependencies used (e.g. `asciidoctor`).
Thus I'd propose to keep plugin configured, enforcing `1.8+`, as above.
And I'd like
```
<message>You need JDK8 or lower</message>
```
to be corrected, as currently it does not reflect rule (which fixes JDK to 1.8x), but also suggest that using earlier JDK is fine (which isn't).

ggam:
@ivargrimstad Jakarta EE 8 targets Java SE 8 for runtime, but some specs already allow compilation with newer Java versions, and in fact most application servers support Java 11 nowadays. I think it's a good thing to allow compiling with latest Java whenever possible.

@pzygielo you're right, thanks. I've already changed it.

pzygielo:
And how about that message displayed on rule violation?
```xml
<message>You need JDK8 or lower</message>
```

ggam:
Embarrasing... Done!

ggam:
The reason is that some people (like me) will have a newer JDK as a
default. With this, there's no need to change versions.

PR #34, opened 2019-07-03
restructure Platform and WebProfile spec creation

kwsutter:
These changes are along the lines of the ones we just merged for [Bill's Spec Template](https://github.com/bshannon/spec-template) project.


kwsutter:
FYI, I split the Platform and WebProfile content into separate directories.  This will allow more granular updates to these documents (eventually).

kwsutter:
> You can't remove the Oracle copyright line. You can add other copyright lines if you want.

Bill, How about if I put it right before the EFSL?  This would be in a similar position as the original Java EE Platform spec for the Oracle copyright.

bshannon:
Ya, that looks fine, thanks.

PR #36, opened 2019-07-04
Update Security section to use Jakarta terms

starksm64:
Signed-off-by: starksm64 <starksm64@gmail.com>

### DIFF
--- a/specification/src/main/asciidoc/platform/Security.adoc
+++ b/specification/src/main/asciidoc/platform/Security.adoc
@@ -139,17 +139,17 @@ the enterprise bean, using the user’s credential to establish a secure
 association between the JSP page and the enterprise bean (as shown in
 _link:#a269[See Invoking an Enterprise
 Bean Business Method]_ ). The association is implemented as two related
-security contexts, one in the web server and one in the EJB container.
+security contexts, one in the web server and one in the Jakarta Enterprise Beans container.
 
 image:Platform_Spec-7.png[image]
 
 === [[a269]]Invoking an Enterprise Bean Business Method
 
-The EJB container is responsible for enforcing
+The Jakarta Enterprise Beans container is responsible for enforcing
 access control on the enterprise bean method. It consults the security
 policy (derived from the deployment descriptor) associated with the
 enterprise bean to determine the security roles that are permitted
-access to the method. For each role, the EJB container uses the security
+access to the method. For each role, the Jakarta Enterprise Beans container uses the security
 context associated with the call to determine if it can map the caller
 to the role.
 
@@ -344,7 +344,7 @@ decisions made by security aware applications. Programmatic security is
 useful when declarative security alone is not sufficient to express the
 security model of the application. The API for programmatic security
 consists of methods of the Jakarta Security _SecurityContext_
-interface, and methods of the EJB _EJBContext_ interface and the servlet
+interface, and methods of the Jakarta Enterprise Beans _EJBContext_ interface and the servlet
 _HttpServletRequest_ interface. The methods of the Jakarta  Security 
 _SecurityContext_ interface are intended to supersede the corresponding
 methods of the _EJBContext_ and _HttpServletRequest_ interfaces.
@@ -387,7 +387,7 @@ association between the components.
 Since a container provides security in Jakarta EE,
 secure associations for a component are typically established by a
 container. Secure associations for web access are specified here. Secure
-associations for access to enterprise beans are described in the EJB
+associations for access to enterprise beans are described in the Jakarta Enterprise Beans
 specification.
 
 Product Providers may allow for control over
@@ -462,7 +462,7 @@ service or a security service.
 === HTTP Login Gateways
 
 Secure interoperability between enterprise
-beans in different security policy domains is addressed in the EJB
+beans in different security policy domains is addressed in the Jakarta Enterprise Beans
 specification. In addition, a component may choose to log in to a
 foreign server via HTTP. An application component can be configured to
 use SSL mutual authentication for security when accessing a remote
@@ -719,14 +719,14 @@ A web container reports that no user has been
 authenticated by returning _null_ from the _SecurityContext_ method
 _getCallerPrincipal_ or the _HttpServletRequest_ method
 _getUserPrincipal_ . This is different than the result of the
-_EJBContext_ method _getCallerPrincipal_ . The EJB specification
+_EJBContext_ method _getCallerPrincipal_ . The Jakarta Enterprise Beans specification
 requires that the _EJBContext_ method _getCallerPrincipal_ always return
 a valid _Principal_ object. This method can never return _null_ . The
 _SecurityContext_ method _getCallerPrincipal_ can also be called in the
-EJB container, and still returns _null_ for anonymous users.
+Jakarta Enterprise Beans container, and still returns _null_ for anonymous users.
 
 In Jakarta EE products that contain both a web
-container and an EJB container, components running in a web container
+container and an Jakarta Enterprise Beans container, components running in a web container
 must be able to call enterprise beans even when no user has been
 authenticated in the web container. When a call is made in such a case
 from a component in a web container to an enterprise bean, a Jakarta EE
@@ -746,10 +746,10 @@ and enterprise bean containers.
 This specification does not specify how a Jakarta
 EE product should choose a principal to represent unauthenticated users,
 although future versions of this specification may add requirements in
-this area. Note that the EJB specification does include requirements in
-this area when using the EJB interoperability protocol. Applications are
+this area. Note that the Jakarta Enterprise Beans specification does include requirements in
+this area when using the Jakarta Enterprise Beans interoperability protocol. Applications are
 encouraged to use the Run As capability in cases where the web component
-may be unauthenticated and needs to call EJB components.
+may be unauthenticated and needs to call Jakarta Enterprise Beans components.
 
 === Application Client User Authentication
 
@@ -893,7 +893,7 @@ Servlet specifications.
 
 === Propagated Caller Identities.
 
-In a Jakarta EE product that contains an EJB
+In a Jakarta EE product that contains an Jakarta Enterprise Beans
 container, it must be possible to configure the Jakarta EE product so that
 a propagated caller identity is used in all authorization decisions.
 With this configuration, for all calls to all enterprise beans from a
@@ -924,7 +924,7 @@ Jakarta EE products must also support the Run As
 capability that allows the Application Component Provider and the
 Deployer to specify an identity under which an enterprise bean or web
 component must run. In this case it is the Run As identity that is
-propagated to subsequent EJB components, rather than the original caller
+propagated to subsequent Jakarta Enterprise Beans components, rather than the original caller
 identity.
 
 Note that this specification doesn’t specify
@@ -952,7 +952,7 @@ deploy applications and components using exactly the mappings and
 assignments specified in their metadata annotations or deployment
 descriptors.
 
-As described in the EJB specification and the
+As described in the Jakarta Enterprise Beans specification and the
 Servlet specification, a Jakarta EE product must provide a deployment tool
 or tools capable of assigning the security roles in metadata annotations
 or deployment descriptors to the entities that are used to determine
--- a/specification/src/main/asciidoc/platform/Security.adoc
+++ b/specification/src/main/asciidoc/platform/Security.adoc
@@ -1,7 +1,7 @@
 == [[a235]]Security
 
 This chapter describes the security
-requirements for the Jakarta™ Platform, Enterprise Edition (Jakarta EE) that
+requirements for the Jakarta™, Enterprise Edition (Jakarta EE) that
 must be satisfied by Jakarta EE products.
 
 In addition to the Jakarta EE requirements, each
@@ -315,7 +315,7 @@ a Jakarta EE environment. A container provides two kinds of security
 Declarative security refers to the means of
 expressing an application’s security structure, including security
 roles, access control, and authentication requirements in
-non-programmatic form. Jakarta language annotations and the deployment
+non-programmatic form. Java language annotations and the deployment
 descriptor are the primary vehicles for declarative security in the Jakarta
 EE platform.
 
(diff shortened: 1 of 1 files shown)


specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -1,11 +1,11 @@
 == [[a235]]Security
 
 This chapter describes the security
-requirements for the Java™ Platform, Enterprise Edition (Java EE) that
-must be satisfied by Java EE products.
+requirements for the Jakarta™ Platform, Enterprise Edition (Jakarta EE) that
```
ivargrimstad: Should we get rid of the TM as well?

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -315,8 +315,8 @@ a Java EE environment. A container provides two kinds of security
 Declarative security refers to the means of
 expressing an application’s security structure, including security
 roles, access control, and authentication requirements in
-non-programmatic form. Java language annotations and the deployment
-descriptor are the primary vehicles for declarative security in the Java
+non-programmatic form. Jakarta language annotations and the deployment
```
ggam: This Java reference should be kept

specification/src/main/asciidoc/platform/Security.adoc
smillidge: Jakarta is a trademark of the Eclipse Foundation see https://www.eclipse.org/legal/logo_guidelines.php

I would follow the rule in the above document `The appropriate trademark symbol (i.e. ™ or ®) must appear at both the first and most prominent use of Eclipse trademarks`

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -343,9 +343,9 @@ Programmatic security refers to security
 decisions made by security aware applications. Programmatic security is
 useful when declarative security alone is not sufficient to express the
 security model of the application. The API for programmatic security
-consists of methods of the Java EE Security API _SecurityContext_
+consists of methods of the Jakarta Security _SecurityContext_
 interface, and methods of the EJB _EJBContext_ interface and the servlet
```
smillidge: Use of EJB here should be changed to Jakarta Enterprise Beans as it is referring to the Jakarta Enterprise Beans api rather than the JCP api

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -384,7 +384,7 @@ confidentiality or integrity.
 . Setting up a security context for the
 association between the components.
 
-Since a container provides security in Java EE,
+Since a container provides security in Jakarta EE,
 secure associations for a component are typically established by a
 container. Secure associations for web access are specified here. Secure
 associations for access to enterprise beans are described in the EJB
```
smillidge: Similar reference to EJB which needs to become Jakarta Enterprise Beans

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -440,7 +440,7 @@ For example,
 
 . A Deployer may have mapped a security role to
 a user group in the operational environment or may depend on the default
-mapping of security roles to user groups as defined by the Java EE
+mapping of security roles to user groups as defined by the Jakarta EE
 Security API specification. In this case, the user group of the calling
```
smillidge: I think this specification may now be called Jakarta Security

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -725,14 +725,14 @@ a valid _Principal_ object. This method can never return _null_ . The
 _SecurityContext_ method _getCallerPrincipal_ can also be called in the
 EJB container, and still returns _null_ for anonymous users.
```
smillidge: Use of EJB

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -725,14 +725,14 @@ a valid _Principal_ object. This method can never return _null_ . The
...
 
-In Java EE products that contain both a web
+In Jakarta EE products that contain both a web
 container and an EJB container, components running in a web container
```
smillidge: Use of EJB

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -743,7 +743,7 @@ server, or per session, or per application.
 choose which principal to use through the Run As capability of the web
 and enterprise bean containers.
 
-This specification does not specify how a Java
+This specification does not specify how a Jakarta
 EE product should choose a principal to represent unauthenticated users,
 although future versions of this specification may add requirements in
 this area. Note that the EJB specification does include requirements in
```
smillidge: Use of EJB

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -861,22 +861,22 @@ authentication requirements can be found in the Connector specification.
...
 products.
 
 === Code Authorization
 
-A Java EE product may restrict the use of
+A Jakarta EE product may restrict the use of
 certain Java SE classes and methods to secure and ensure proper
-operation of the system. The minimum set of permissions that a Java EE
+operation of the system. The minimum set of permissions that a Jakarta EE
 product is required to grant to a Java EE application is defined in
```
smillidge: Should be Jakarta

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -885,19 +885,19 @@ permissions that a component needs for access.
 
 === Caller Authorization
 
-A Java EE product must enforce the access
+A Jakarta EE product must enforce the access
 control rules specified at deployment time (see
 link:#a451[See Deployment
 Requirements]”) and more fully described in the EJB and Servlet
```
smillidge: Use of EJB

smillidge: You may want to refer to Jakarta Servlet but I'm not sure

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -885,19 +885,19 @@ permissions that a component needs for access.
...
 specifications.
 
 === Propagated Caller Identities.
 
-In a Java EE product that contains an EJB
-container, it must be possible to configure the Java EE product so that
+In a Jakarta EE product that contains an EJB
```
smillidge: Use of EJB

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -930,29 +930,29 @@ identity.
...
+such as files. However, the Jakarta Authorization
 specification does specify the relationship between the Run As identity
 and the access control context used by the Java SE security manager.
 
 === [[a451]]Deployment Requirements
 
-All Java EE products must implement the access
+All Jakarta EE products must implement the access
 control semantics described in all included component specifications,
 such as the EJB, JSP, and Servlet specifications; provide a means of
```
smillidge: Use of EJB and maybe the other names should refer to their Jakarta spec names

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -930,29 +930,29 @@ identity.
...
-While most Java EE products will allow the
+While most Jakarta EE products will allow the
 Deployer to customize the role mappings and change the assignment of
-roles to methods, all Java EE products must support the ability to
+roles to methods, all Jakarta EE products must support the ability to
 deploy applications and components using exactly the mappings and
 assignments specified in their metadata annotations or deployment
 descriptors.
 
 As described in the EJB specification and the
```
smillidge: Use of EJB

specification/src/main/asciidoc/platform/Security.adoc
ivargrimstad: Right

specification/src/main/asciidoc/platform/Security.adoc
kwsutter: I think we should drop the use of "Platform" in this context since it was referring to the "Java Platform", not the Platform spec.  We should just call this "Jakarta Enterprise Edition".

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -44,16 +44,16 @@ tamper-resistant record of security related events for the purpose of
...
 being considered for future versions of this specification are briefly
 mentioned in link:#a457[See Future
 Directions]”.
 
 === A Simple Example
 
-The security behavior of a Java EE environment
+The security behavior of a Jakarta EE environment
 may be better understood by examining what happens in a simple
 application with a web client, a JSP user interface, and enterprise bean
```
kwsutter: Change JSP to either Jakarta Server Pages, or just go generic like "server pages" (similar to the generic reference to "enterrpise bean business logic".)

specification/src/main/asciidoc/platform/Security.adoc
kwsutter: Nice catch, @ggam!

specification/src/main/asciidoc/platform/Security.adoc
kwsutter: I would just combine the two -- "..described in the Jakarta Enterprise Beans and Servlet specifications"

specification/src/main/asciidoc/platform/Security.adoc
pplupo: When you changed Java EE for Jakarta EE you missed the "EE". Isn't it supposed to be "Jakarta EE Security..."?

specification/src/main/asciidoc/platform/Security.adoc
bshannon: Is Jakarta EE a platform or not?  (And what exactly makes something a "platform"?)
And would we expect other "editions" of the "Jakarta Platform"?

We should be clear on whether we think of this as a platform, but even so that
doesn't require "platform" to be part of the name so I agree with removing it here.

specification/src/main/asciidoc/platform/Security.adoc
bshannon: Using JSP is fine, as long as you include the appropriate disclaimer at the beginning of the spec.
I suspect we're going to want to use "JSP" elsewhere in the spec as well.

specification/src/main/asciidoc/platform/Security.adoc
bshannon: No, it's just "Jakarta Security".

Using "EJB" is fine, as long as the disclaimer is added to the beginning of the spec.

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -769,7 +769,7 @@ interface for interactions with the user to gather authentication data.
 In addition, an application client may provide a class that implements
 the _javax.security.auth.callback.CallbackHandler_ interface and specify
 the class name in its deployment descriptor (see
-link:#a3399[See Java EE Application
+link:#a3399[See Jakarta EE Application
```
bshannon: Do these cross-chapter links still work after splitting the single document into multiple documents?

kwsutter: Not sure, to be honest, @bshannon.  Haven't worried about it yet...  I'm sure we can make them work somehow.  Or, we can re-combine the sections if we have to...  My first goal is to just to get a new "Jakarta" baseline...

kwsutter: @bshannon Turns out that it's no worse than it was ...  :-(  I just went back to the original single adoc version of the spec and the none of the hotlinks in the pdf work.  Same error with the old format as with the new format...  So, we need to figure out the proper reference syntax and do some more updating...  This is so much fun...

bshannon: Yes, there's a separate problem with the links in the pdf.  If someone could figure out
how to make that work that would be great.  But the links *did* work in the html.

kwsutter: Just starting to look into this link issue...  Some of the links seem to have direct links to a corresponding .html page...  That's why many of them work in the generated html...  I'll work on a couple of updates, create a new PR, and get feedback before trying to update all of the references.

starksm64:
I have updated everything but the links as it is not clear what they should be changed to.

kwsutter:
Forget about the links.  We'll get that cleaned up across all of the sections of the spec.

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -1,7 +1,7 @@
 == [[a235]]Security
 
 This chapter describes the security
-requirements for the Jakarta™ Platform, Enterprise Edition (Jakarta EE) that
+requirements for the Jakarta™, Enterprise Edition (Jakarta EE) that
```
kwsutter: Drop the comma after Jakarta (TM).

specification/src/main/asciidoc/platform/Security.adoc
smillidge: You still have the EJB reference

specification/src/main/asciidoc/platform/Security.adoc
```diff
@@ -108,7 +108,7 @@ each role to determine if it can map the user to the role.
 _link:#a261[See URL Authorization]_
 shows this process.
 
-image:Platform_Spec-5.png[image]
+image:images/Platform_Spec-5.png[image]
```
bshannon: There seems to be some disagreement over how to specify the images.
This all worked for me without any path information in the image references,
for both html and pdf.  I don't understand why "images/" needs to be added
here.  And I've seen "../images/" added in other sections.

Someone needs to explain why it doesn't work as is, then explain how to
update it consistently in all sections.

kwsutter: @bshannon All of the image processing still needs work.  Let's not hold up this review on this global item.  This section is getting too far out of date with regards to the other cleanup.  Can we approve this PR and log an issue for images?  It's on my list of TODOs, but an issue would be even better reminder.  Thanks.

starksm64:
Resolving the conflicts with changes made for links has introduced an ECA error.

specification/src/main/asciidoc/platform/Security.adoc
bshannon: I guess I would prefer that this just be put back to the way it was
and then if there really is something broken they can all be fixed together.
But I won't hold up the review for this.

kwsutter:
> Resolving the conflicts with changes made for links has introduced an ECA error.

Yeah, sometimes if you attempt to resolve the merge conflicts via the gui, then it would do a commit with an anonymous id and then you get the ECA failure...  I think you can just do the signed-off processing and it should clear up.

starksm64:
Replaced by #71

PR #41, opened 2019-07-04
First cut FutureDirections.adoc for Jakarta EE 8

smillidge:
Signed-off-by: smillidge <steve.millidge@payara.fish>

### DIFF
--- a/specification/src/main/asciidoc/platform/FutureDirections.adoc
+++ b/specification/src/main/asciidoc/platform/FutureDirections.adoc
@@ -1,7 +1,6 @@
 == [[a3439]]Future Directions
 
-This version of the Java™ Platform,
-Enterprise Edition (Java EE) specification includes most of the
+This version of the Jakarta™ EE Platform specification includes most of the
 facilities needed by enterprise applications. Still, there is always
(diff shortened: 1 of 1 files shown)


specification/src/main/asciidoc/platform/FutureDirections.adoc
```diff
@@ -1,27 +1,26 @@
...
+=== Jakarta EE SPI
 
-Many of the APIs that make up the Java EE
+Many of the APIs that make up the Jakarta EE
 platform include an SPI layer that allows service providers or other
 system level components to be plugged in. This specification does not
 describe the execution environment for all such service providers, nor
 the packaging and deployment requirements for all service providers.
-However, the Java EE Connector Architecture does define the requirements
+However, the Jakarta Connectors Specification does define the requirements
```
xstefank: It seems that it should be "Jakarta Connectors specification" with lower case specification if I'm not missing something.

kwsutter: I think either one would suffice in this case.  But, the capitalized Specification is probably most correct.

PR #47, opened 2019-07-05
Update SecurityProviderInterface.adoc to use Jakarta terms

xstefank:
Signed-off-by: xstefank <xstefank122@gmail.com>

### DIFF
--- a/specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
+++ b/specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
@@ -1,64 +1,61 @@
 == Service Provider Interface
 
-The Java™ Platform, Enterprise Edition (Java
-EE) includes several technologies that are primarily intended to be used
-to extend the capabilities of the Java EE containers. In addition, some
-Java EE technologies include service provider interfaces along with
-their application programming interfaces. A Java EE profile may include
-some or all of these facilities, as described in
-link:#a3212[See Profiles]”.
+The Jakarta™ EE Platform includes several technologies that are primarily
+intended to be used to extend the capabilities of the Jakarta EE containers.
+In addition, some Jakarta EE technologies include service provider interfaces
+along with their application programming interfaces. A Jakarta EE profile may
+include some or all of these facilities, as described in link:#a3212[See Profiles]”.
 
-=== Java™ EE Connector Architecture
+=== Jakarta™ Connectors
 
 The Connector API defines how resource adapters
-are packaged and integrated with any Java EE product. Many types of
+are packaged and integrated with any Jakarta EE product. Many types of
 service providers can be provided using the Connector API and packaging,
-including JDBC drivers, JMS providers, and JAXR providers. All Java EE
-products must support the Connector APIs, as specified in the Connector
-specification.
+including JDBC drivers, Jakarta Messaging providers, and Jakarta XML Registries
+providers. All Jakarta EE products must support the Connector APIs, as specified
+in the Connector specification.
 
-The Connector specification is available at
-_http://jcp.org/en/jsr/detail?id=322_ .
+=== Jakarta™ Authorization
 
-=== Java™ Authorization Service Provider Contract for Containers
+The Jakarta Authorization specification defines the contract
+between a Jakarta EE container and an authorization policy provider.
 
-The JACC specification defines the contract
-between a Java EE container and an authorization policy provider.
+=== Jakarta™ Transactions
 
-The JACC specification is available at
-_http://jcp.org/en/jsr/detail?id=115_ .
-
-=== Java™ Transaction API
-
-The Java Transaction API defines the
+The Jakarta Transactions defines the
 _TransactionSynchronizationRegistry_ interface that is intended for use
 by system level application server components such as persistence
-managers, resource adapters, as well as EJB and Web application
+managers, resource adapters, as well as Jakarta Enterprise Beans and Web application
 components. This provides the ability to register synchronization
 objects with special ordering semantics, associate resource objects with
 the current transaction, get the transaction context of the current
 transaction, get current transaction status, and mark the current
 transaction for rollback.
 
-The JTA specification is available at
-_http://jcp.org/en/jsr/detail?id=907_ .
+=== Jakarta™ Persistence
 
-=== Java™ Persistence
-
-Java Persistence provides interfaces in the
+Jakarta Persistence provides interfaces in the
 _javax.persistence.spi_ package that allow a persistence provider to be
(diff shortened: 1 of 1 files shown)


specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
```diff
@@ -1,64 +1,62 @@
...
-some or all of these facilities, as described in
-link:#a3212[See Profiles]”.
+The Jakarta™ EE Platform includes several technologies that are primarily
+intended to be used to extend the capabilities of the Jakarta EE containers.
+In addition, some Jakarta EE technologies include service provider interfaces
+along with their application programming interfaces. A Jakarta EE profile may
+include some or all of these facilities, as described in link:#a3212[See Profiles]”.
 
-=== Java™ EE Connector Architecture
+=== Jakarta™ EE Connectors
```
kazumura: should be  Jakarta™ Connectors

xstefank: Thanks. Fixed.

specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
```diff
@@ -1,64 +1,62 @@
...
-=== Java™ Authorization Service Provider Contract for Containers
+=== Jakarta™ Authorization
 
-The JACC specification defines the contract
-between a Java EE container and an authorization policy provider.
+The Jakarta Authorization specification defines the contract
+between a Jakarta EE container and an authorization policy provider.
 
-The JACC specification is available at
+The Jakarta Authorization specification is available at
```
kwsutter: This link is actually for the old JACC specification...  Since we don't have the updated Jakarta Authorization Spec yet, we should probably revert this change to explain that the original JACC spec is available here...  Either that or just remove this reference in this portion of the doc.

specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
```diff
@@ -1,64 +1,62 @@
...
-managers, resource adapters, as well as EJB and Web application
+managers, resource adapters, as well as Jakarta Enterprise Beans and Web application
 components. This provides the ability to register synchronization
 objects with special ordering semantics, associate resource objects with
 the current transaction, get the transaction context of the current
 transaction, get current transaction status, and mark the current
 transaction for rollback.
 
-The JTA specification is available at
+The Jakarta Transactions specification is available at
```
kwsutter: Same comment here...  This link is the Java EE Transactions doc, not the Jakarta Spec...  We should have a single reference section at the end of the doc with references to the old and new Specs.  Maybe we should just be pointing at that?  Or, just remove this explicit reference.

specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
```diff
@@ -1,64 +1,62 @@
...
+=== Jakarta™ Persistence
 
-Java Persistence provides interfaces in the
+Jakarta Persistence provides interfaces in the
 _javax.persistence.spi_ package that allow a persistence provider to be
-plugged into the Java Persistence framework.
+plugged into the Jakarta Persistence framework.
 
-The Java Persistence specification is available
+The Jakarta Persistence specification is available
```
kwsutter: Same comment again...

specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
```diff
@@ -1,64 +1,62 @@
...
-can be discovered and used through the JavaMail API. This allows the
-JavaMail API to be extended with support for new mail protocols and
+The Jakarta Mail specification describes how
+Jakarta Mail protocol providers can be packaged and distributed so that they
+can be discovered and used through the Jakarta Mail API. This allows the
+Jakarta Mail API to be extended with support for new mail protocols and
 mailbox formats.
 
-The JavaMail specification is available at
+The Jakarta Mail specification is available at
```
kwsutter: And, again...

specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
xstefank: I agree. Can you point me to the new spec documents, please? I can't find them.

xstefank:
@kwsutter I moved JCP links to the end of the documents and added section also for new Jakarta spec documents. I will make a note to update this section when the documents will be available.

xstefank:
@kwsutter can you take a long again, please?

specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
kwsutter: We don't have the "standard" location for the updates specs yet...  At this point, let's just be consistent.  If we want to reference the JCP spec, then that's okay.  But, we shouldn't talk about Jakarta Spec and then point at the JCP location.

kwsutter:
@xstefank Thanks, Martin.  That's a good compromise by moving the references to the end of the section.  If or when we have the proper specs for Jakarta, we can update this section.  Thanks.  I will approve and merge.

xstefank:
Thanks @kwsutter!

specification/src/main/asciidoc/platform/ServiceProviderInterface.adoc
```diff
@@ -1,64 +1,61 @@
...
+Jakarta Mail protocol providers can be packaged and distributed so that they
+can be discovered and used through the Jakarta Mail API. This allows the
+Jakarta Mail API to be extended with support for new mail protocols and
 mailbox formats.
 
-The JavaMail specification is available at
-_http://jcp.org/en/jsr/detail?id=919_ .
+=== Specification documents
+
+==== Original Java™ EE specification documents
```
bshannon: It will be cleaner and less confusing if we don't reference the JCP specs at all.
It would be better to omit these sections at the end and just depend on the
appendix that will link to all the specification documents.

xstefank: @bshannon ok, but I believe there are more sections that reference original JCP links. It would be better to do this in one PR for all documents after the new spec links are available.

bshannon: The expected form of spec links is described in [this message from Wayne](https://www.eclipse.org/lists/jakarta.ee-spec/msg00035.html).

xstefank: Would then make sense to update it right away? I saw the message after the PR was merged.

PR #56, opened 2019-07-08
Update TM syntax in asciidoc

kwsutter:
@smillidge, the %trade; syntax didn't render correctly for the TM reference in the Compatibility section.  I corrected this to just use (TM).  It looks like I dragged along some extra "empty" commits as I was getting my fork up-to-date, but you'll see that the only code change was to the Trademark reference.  Thanks.

### DIFF
--- a/specification/src/main/asciidoc/platform/CompatibilityMigration.adoc
+++ b/specification/src/main/asciidoc/platform/CompatibilityMigration.adoc
@@ -1,7 +1,7 @@
 == Compatibility and Migration
 
 This chapter summarizes compatibility and
-migration issues for the Jakarta&trade; EE platform. The specifications for each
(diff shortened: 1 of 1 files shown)


specification/src/main/asciidoc/platform/CompatibilityMigration.adoc
```diff
@@ -1,7 +1,7 @@
 == Compatibility and Migration
 
 This chapter summarizes compatibility and
-migration issues for the Jakarta&trade; EE platform. The specifications for each
+migration issues for the Jakarta(TM) EE platform. The specifications for each
```
Emily-Jiang: Why does not use Jakarta™?

kwsutter: Either approach works...  The current approach of using \&trade; does not...

PR #62, opened 2019-07-11
create skeletal managed beans spec

kwsutter:
Until we get the real thing, here's the skeletal ManagedBeans spec based on @bshannon's spec-template.

### DIFF
--- a/specification/src/main/asciidoc/managedbeans/Introduction.adoc
+++ b/specification/src/main/asciidoc/managedbeans/Introduction.adoc
@@ -1,0 +1,29 @@
+== Introduction
+
+This specification defines Managed Beans for the Jakarta(TM) EE Platform.
+
+=== What are Managed Beans?
+
+Managed Beans are container-managed objects with minimal requirements, otherwise known under the acronym “POJOs” (Plain Old Java Objects).
+They support a small set of basic services, such as resource injection, lifecycle callbacks, and interceptors.
+Other, more advanced, aspects will be introduced in companion specifications, so as to keep the basic model as simple and as universally useful as possible.
+
+=== Why Managed Beans?
+
+Managed Beans offer a lightweight component model aligned with the rest of the Jakarta EE Platform.
+
+By supporting the common resource injection and lifecycle services, Managed Beans fit well into the Jakarta EE programming model.
+At the same time, their lightweight nature makes them a natural starting point to encapsulate application functionality, with the knowledge that they can be morphed into more powerful components if and when the need occurs.
+In this sense, they can be seen as a Jakarta EE platform-enhanced version of the JavaBeans component model found on the Java(TM) SE platform.
+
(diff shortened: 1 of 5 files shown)


specification/src/main/asciidoc/managedbeans/Introduction.adoc
```diff
@@ -0,0 +1,29 @@
+== Introduction
+
+This specification defines Managed Beans for the Jakarta(TM) EE Platform.
+
+=== What are Managed Beans?
```
bshannon: This looks like a copy and paste from the JCP Managed Beans spec, which we haven't been
given permission to use.  I think all the following content needs to be removed.

specification/src/main/asciidoc/managedbeans/Introduction.adoc
```diff
@@ -0,0 +1,29 @@
+== Introduction
+
+This specification defines Managed Beans for the Jakarta(TM) EE Platform.
```
bshannon: This should be a copy of the approved scope statement:

> The Jakarta Managed Beans specification defines a set of basic services for container-managed objects
with minimal requirements, otherwise known under the acronym “POJOs” (Plain Old Java Objects).

(Reword as appropriate.)

specification/src/main/asciidoc/managedbeans/Introduction.adoc
kwsutter: Yes, it is a copy-and-paste...  I really didn't think that this front matter would be a problem.  It's what our scope statement is based off of...  But, we think it will be a problem, then I can skinny it down to just the Scope statement.

specification/src/main/asciidoc/managedbeans/Introduction.adoc
kwsutter: I looked for a public Managed Beans Scope statement, but didn't find one on our project pages...  Thanks for finding this in Issue #8.  Again, if we think this is the only content allowed for the skeletal spec, I'll narrow the wording.  Thanks!

specification/src/main/asciidoc/managedbeans/Introduction.adoc
bshannon: It's obviously copyright infringement...

specification/src/main/asciidoc/managedbeans/Introduction.adoc
bshannon: Sadly, yes, it's the only content expected.

specification/src/main/asciidoc/managedbeans/Introduction.adoc
kwsutter: Okay, will remove...

specification/src/main/asciidoc/managedbeans/Introduction.adoc
kwsutter: Yep, pretty sad...  Watch for the next commit...

PR #67, opened 2019-07-13
Required Section Updates

kwsutter:
This still isn't perfect, but it's 10x better than it was...  This correction also "found" a missing section (ApplicationProgrammingInterface) that was hidden in the original contribution.  Personally, I wouldn't spend a great deal of time reviewing this.  If you apply this PR and take a look at the TOC, you'll see the benefits.

### DIFF
--- a/specification/src/main/asciidoc/Platform-spec.adoc
+++ b/specification/src/main/asciidoc/Platform-spec.adoc
@@ -10,7 +10,7 @@
 :license: Eclipse Foundation Specification License v1.0
 :source-highlighter: coderay
 :toc: left
-:toclevels: 4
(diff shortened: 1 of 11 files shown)


kwsutter:
I'm hoping to get a quick approval just so that I can continue to make progress on getting the specs readable...  Since I'm touching just about every section with this PR, it would be good to just merge this before piling on more changes.  This really could be considered part of the original contribution since I'm just fixing errors with the sections.  We technically don't need an approval on this github repo before merging, but it still would be nice...  Thanks.

kwsutter:
> Greetings from the beaches of Crete 🌞

Thank you, Ivar!

PR #68, opened 2019-07-15
Update Spec documents links and refs

kwsutter:


kwsutter:
I did not do all of these changes by hand, @bshannon...  :-)

I used the following regex find/replace strings in Atom and did all of the adoc files in the Platform and Web Profile specs.  I need to find a better place for this information (wiki?) so that the future specs from Oracle can benefit.

**Change all of the References to work with both PDF and HTML generation**
Find:  link\:[^#]*#(\w+)\[See\s([^\]]+)\]

Replace:  <<$1, $2>>

**Move the links defined as part of the section headings so that the Refs wouldn't take us to the TOC**
Find:  (\={2,5})\s(\[\[\w+\]\])\s*([^\n]+)

Replace:  $2\n$1 $3

Again, a Review might be tedious...  Looking for a cursory review and Approval just so that we can get a more readable document.  There is still work to do....  Thanks!

kwsutter:
Thanks for the Approval, David!  Moving forward...

PR #72, opened 2019-07-17
More section updates

kwsutter:
Found another "missing" section due to bad asciidoc in the original contribution.  Continued to update the section headings to match the TOC of the Java EE 8 doc.  And, added the [appendix] tag on the three chapters that were Appendices in the original Java EE 8 spec.

Once again, looking for a quick review and approval just so that we can continue to make progress.  Thanks!


kwsutter:
Thanks, Bill.  The Framemaker -> Asciidoc tools that were used did an okay job at the 10,000 foot level, but there are so many things that have required manual detection and intervention.  We're getting closer, but it's still not ready for review.  Thanks for the quick approval!

Issue #75, opened 2019-07-17
Clean up Tables in Platform Spec


kwsutter:
Tables are not getting formatted at all.  I've started on this issue.  So far, haven't figured out a nice pattern to script this, but working on it.

kwsutter:
Resolved via PR #78

Issue #77, opened 2019-07-17
Footnotes are not formatted or referenced correctly


kwsutter:
Footnotes are currently all put into a single location at the end of the generated doc, with no means of easily getting back to the text you were reading.  And, the references are not superscripted.  It would be best if the footnotes could be generated on the same page.  And, use superscripts.  At the very least, they need to be formatted better than they are.

kwsutter:
Comment left by @bshannon on PR #78...

Generally looks fine. One comment below. Also, consider using this for footnotes:
https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#section-footnotes
https://asciidoctor.org/docs/user-manual/#user-footnotes

kwsutter:
Footnotes are kind of ugly with the pdf and html generation from the asciidoc.  The only way I could them to work is to declare them inline.  The end result looks quite good, but the asciidoc isn't the prettiest.  It seems that the asciidoctor team continues to work on improving this support.  I had to update which version of the maven plugin to use to get the nice output we have now.  Eventually, they may allow the use of externalized footnotes.  But, for now we'll have to live with the inlined version.

kwsutter:
Resolved via PR #83.

PR #78, opened 2019-07-17
tables, images, and more links/refs

kwsutter:
Tables were cleaned up for Issue #75.  Also cleaned up Images.  More missing links and references.  And, some more misc cleanup.  Cursory review and approval requested.  Thank you!

### DIFF
--- a/specification/src/main/asciidoc/platform/RevisionHistory.adoc
+++ b/specification/src/main/asciidoc/platform/RevisionHistory.adoc
@@ -7,10 +7,10 @@
 
 * Java EE 8 requires Java SE 8.
 * Added
-<<a2713, Java API for JSON Binding 1.0 (JSON-B) Requirements>>.”
-* Added link:api.html#UNKNOWN[].”
+<<a2713, Java API for JSON Binding 1.0 (JSON-B) Requirements>>.
+* Added "Model-View-Controller (MVC) 1.0 Requirements”
 * Added
-<<a149, Java EE Product Packaging>>”; updated <<a3153, Deployment>>.”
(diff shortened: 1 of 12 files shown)


specification/src/main/asciidoc/platform/RevisionHistory.adoc
```diff
@@ -7,10 +7,10 @@
 
 * Java EE 8 requires Java SE 8.
 * Added
```
bshannon: The revision history should be cleared out and replaced for Jakarta EE 8.

kwsutter:
> Generally looks fine. One comment below. Also, consider using this for footnotes:
> https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#section-footnotes
> https://asciidoctor.org/docs/user-manual/#user-footnotes

Thanks.  The footnote work will be covered by this Issue #77.  It was next on my list...

specification/src/main/asciidoc/platform/RevisionHistory.adoc
kwsutter: I the Revision History does show the history before Jakarta...  I'd rather leave it.

bshannon: Didn't we remove all the acknowledgements for those versions?

As it is, the revision history is only for the lifetime of the Java EE 8 spec.
We clear out the revision history with each new version.  Seems like we
should do the same here, indicating how this version of this Jakarta EE
spec came to be.  The things that happened early in the Java EE 8 spec
development are irrelevant.

Issue #86, opened 2019-07-20
Update README to increase the min level of maven


kwsutter:
Per this late comment on PR #83, we need to update the README to indicate a slightly newer version of maven is required as a minimum.
https://github.com/eclipse-ee4j/jakartaee-platform/pull/83#issuecomment-513442224

PR #87, opened 2019-07-20
Update and enforce min maven version

pzygielo:
I took liberty to convert prerequisite from readme to enforcer rule, like it's done for JDK.


kwsutter:
This PR resolves Issue #86.

Issue #76, opened 2019-07-17
Code Blocks need cleanup in Platform spec


kwsutter:
The code blocks are not formatted correctly in the Platform spec. Hopefully, this will be an easy fixup.  I'm just documenting all of the items I am finding in case someone else can pick up these changes.

NottyCode:
I volunteer

keilw:
Shouldn't there be `.gitattributes` and `.editorconfig` files for that?

kwsutter:
@keilw This is for the code samples that are embedded in the Platform Specification.  Look at the referenced PR #88.  It's not "real code", just blocks of sample code in the doc,.

kwsutter:
This was resolved via PR #88.

PR #88, opened 2019-07-20
Update code example formatting

NottyCode:
Gone though the code examples to reformat them properly for asciidoc as per #76 . This has consisted of the following:

1. surround in ----
2. Add [source,java] and [source,xml] where relevant
3. Removed _ italic surrounds where not necessary
4. Fixed a few bullet lists which were missing the *
5. Removed some superflous spaces.
6. Remove escape character before {
7.  Used [subs=+quotes] in one case so italic formatting could be used to replicate original spec doc
8. Removed some text at the start of ResourceNamingInjection.adoc that prevented it being rendered into the spec.

### DIFF
--- a/specification/src/main/asciidoc/platform/ResourcesNamingInjection.adoc
+++ b/specification/src/main/asciidoc/platform/ResourcesNamingInjection.adoc
@@ -1,4 +1,4 @@
-Jakarta Enterprise Beans[[a567]]
+[[a567]]
 == Resources, Naming, and Injection
 
 This chapter describes how applications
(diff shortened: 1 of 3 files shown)


specification/src/main/asciidoc/platform/ResourcesNamingInjection.adoc
```diff
@@ -5159,5 +4471,3 @@ Containers may optimize the steps above,
 e.g., by avoiding calls to the actual CDI SPI and relying on
 container-specific interfaces instead, as long as the outcome is the
 same.
-
-'''
```
kwsutter: Do not remove this ''' for generating a line...  This provides a break between the text and the footnotes in this section when generating the pdf.  I guess I should have put a comment indicating the need for this.  Could you just add something along the lines of // generates a line between text and footnotes for pdf and html generation.

Could you also add a similar comment line at the end of ApplicationProgrammingInterface (since you are updating that file as well)?

Thanks!

kwsutter:
This looks great, btw, @NottyCode!  Thanks for doing this.  It looks like you had to touch almost the whole ResourceNamingInjection file!

NottyCode:
@kwsutter I've addressed your comments.

PR #93, opened 2019-07-25
It's Jakarta EE 8, not 8.0.

bshannon:


keilw:
If it does not violate general Eclipse [Version Numbering](https://wiki.eclipse.org/Version_Numbering) (there it's even 3 digits, not 1 or 2) why not.
However, why did the version number of Jakarta Injection API change from "1" to "1.0" in https://github.com/eclipse-ee4j/injection-api/blob/master/pom.xml for a spec that has not changed since Java EE 6?

kwsutter:
@keilw As I mentioned above, let's just go with the single digit release for now.  It's consistent with Java EE 8.  If necessary, we can always extend the digits later, if necessary.  Thanks.

keilw:
@kwsutter As this primarily looks like a parent, that does not define modules which may change, it seems suffienent for now.

PR #97, opened 2019-08-07
update specs per comments received

kwsutter:

Renaming spec file names to be all lowercase to match the directory names and the requested format in to the templates.


kwsutter:
Nice catch, Alasdair!  I definitely only wanted to do a rename.  I'll just close out this PR and start over.

PR #123, opened 2019-10-15
Add meeting minutes to sidebar

ivargrimstad:
Depends on https://github.com/jakartaee/jekyll-theme-jakarta-ee/pull/20 in order to be visible on the webpage, but should be safe to merge anyway.


ivargrimstad:
I'm not sure how formal we need to be. I'll add the other project leads that were present as reviewers as well. Then maybe approve by lazy consensus?

edbratt:
I did manage to join, though quite late and certainly after any 
roll-call had concluded (I don't care if I'm not listed).

I am all for making these types of material available sooner, rather 
than later. Perhaps these could be called "Notes" or "Discussion Recap," 
if "Minutes" are perceived to be more formal. No doubt there will be 
things we will formally want to vote on.

Should the comment about Slack, clarify which channel in the Eclipse 
Foundation workspace? It seems that Slack works better if a member sends 
invitations -- OR, we can ask our Slack Admins to create a link that we 
can share (I'd have to ask Tanja, who that is). (I'm a pretty challenged 
Slacker so maybe there are other options.)

bshannon:
We could also publish the minutes and mark them DRAFT until they are approved in a subsequent meeting.

ivargrimstad:
@edbratt I think it was Tanja that set up the slack workspace. I will ask her if she can provide a link.

@bshannon I'll update the minutes in this PR with DRAFT so we can get them publicly available and approve on next meeting.

PR #128, opened 2019-11-20
Minutes 2019 11 19

ivargrimstad:


minutes/minutes.md
```diff
@@ -2,3 +2,4 @@
 
 * [2019-10-15](2019-10-15.md)
 * [2019-11-12](2019-11-12.md)
+* [2019-11-19](2019-11-12.md)
```
ggam: Wrong link

minutes/2019-11-19.md
```diff
@@ -0,0 +1,105 @@
...
+* Jonathan Gallimore (Tomitribe)
+
+## Agenda
+
+### [KWS] Java SE 8 or Java SE 11 as the minimum. 
+
+Another key decision to allow for continued development efforts.
+* [KWS] Will need to propose a vote on the mailing list indicating the recommendation of Java SE 8 as the minimum, but we would support both Java SE 8 and 11.  (Steve “volunteered” to post the vote.)
+
+### [IG] “Jakartify” the specification documents. Status: https://github.com/eclipse-ee4j/jakartaee-platform/projects/2
```
kwsutter: The Status shouldn't be part of the heading.

minutes/minutes.md
kwsutter: Should we reverse the order of this list so that the most recent is that top of the list?  I think that would be more intuitive, especially as we get more and more of these minutes archived.

ivargrimstad: fixed

minutes/2019-11-19.md
ivargrimstad: fixed

PR #153, opened 2020-02-03
Initial namespace conversion

smillidge:
Signed-off-by: Steve Millidge <steve.millidge@payara.fish>

### DIFF
--- a/specification/pom.xml
+++ b/specification/pom.xml
@@ -27,7 +27,7 @@
     <groupId>jakarta.platform</groupId>
     <artifactId>platform-spec</artifactId>
     <packaging>pom</packaging>
-    <version>9.0-SNAPSHOT</version>
+    <version>9-SNAPSHOT</version>
     <name>Jakarta EE Platform Specification</name>
 
     <properties>
--- a/specification/pom.xml
+++ b/specification/pom.xml
@@ -27,7 +27,7 @@
     <groupId>jakarta.platform</groupId>
     <artifactId>platform-spec</artifactId>
     <packaging>pom</packaging>
-    <version>9-SNAPSHOT</version>
+    <version>9.0-SNAPSHOT</version>
     <name>Jakarta EE Platform Specification</name>
 
     <properties>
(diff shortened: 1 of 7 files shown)


specification/pom.xml
```diff
@@ -27,7 +27,7 @@
     <groupId>jakarta.platform</groupId>
     <artifactId>platform-spec</artifactId>
     <packaging>pom</packaging>
-    <version>9-SNAPSHOT</version>
+    <version>9.0-SNAPSHOT</version>
```
kwsutter: We decided with Jakarta EE 8 that the platform specification (and web profile) will use a single digit for the release version.  Thus, this should stay at 9-SNAPSHOT.

kwsutter:
This looks quite complete, @smillidge!  Could we just revert that pom update and then we could go ahead with approval and merge.  Thanks!

kwsutter:
> This looks good, but you need to verify that any APIs that use property names starting with
> "javax" are actually changing the property names to start with "jakarta". I would expect them
> to, but some may not have been thinking about this issue.
> 
> (And of course this is another issue to be dealt with by any backward compatibility solution.)

Good point, @bshannon.  Actually, I would expect most properties to stay with the same javax prefix.

bshannon:
> Actually, I would expect most properties to stay with the original java or javax prefix.

Really? Why is that?  Sounds like an issue we should discuss and put out a recommendation on.

> I don't think we have to know the "final answer" before merging these changes. Maybe we can create a separate issue to come back and verify the property names before we go with the final version of the Specs...

There were a few instances of property names being changed, I believe.  We don't need to know the
final answer as long as we remember to update this as necessary once the final answer is known.

scottmarlow:
> There were a few instances of property names being changed, I believe.

[One example of a property change to jakarta is in javadoc here.](https://github.com/eclipse-ee4j/jpa-api/blob/master/api/src/main/java/jakarta/persistence/CacheStoreMode.java#L21)

kwsutter:
> > Actually, I would expect most properties to stay with the original java or javax prefix.
> 
> Really? Why is that? Sounds like an issue we should discuss and put out a recommendation on.

When the required javax->jakarta name change was being discussed, the property names did not have to change.  They could be changed, but there was no requirement.  But, I also agree that this item should be discussed.  I'll add it to the Platform agenda for today.

bshannon:
The property names were discussed and approved so I think we're ready to update this.

smillidge:
OK @kwsutter I reverted the pom change

PR #178, opened 2020-04-23
remove the 'Java Compatible' requirement

kwsutter:

Removing the requirement to execute with a "Java Compatible" runtime, which precludes the use of other JVMs such as OpenJ9.  Hopefully, removing just the "Compatible(TM)" designation is sufficient.  Will look for help from our Oracle friends to help verify.

I also found a miss in Jakarta EE 8 with a reference to "Java EE", when it should have been "Jakarta EE".  So, I fixed that as well with this PR.

### DIFF
--- a/specification/src/main/asciidoc/platform/PlatformOverview.adoc
+++ b/specification/src/main/asciidoc/platform/PlatformOverview.adoc
@@ -6,7 +6,7 @@ Jakarta™ EE Platform.
 === Architecture
 
 The required relationships of architectural
-elements of the Java EE platform are shown in
+elements of the Jakarta EE platform are shown in
 _<<a45, Jakarta EE Architecture Diagram>>_.
 Note that this figure shows the logical relationships of the
 elements; it is not meant to imply a physical partitioning of the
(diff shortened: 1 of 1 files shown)


kwsutter:
> As I said in email, if you want to build something that's not compatible, simply don't claim
> that it' compatible. But a compatible Jakarta EE runtime ought to execute on a compatible
> Java SE runtime. I see no reason to remove this requirement.

For reference, the email that Bill referenced is on the [Platform mailing list](https://www.eclipse.org/lists/jakartaee-platform-dev/msg01697.html).  I have posted my reasons for this change in a reply to the mailing list.  In a nutshell, as you can see by my proposed change, I am not removing the requirement of the Java Spec.  I just want to remove the requirement of passing the Java TCK and being labeled as "Java Compatible(TM)", which is not readily available in the open-source world.

NottyCode:
@kwsutter I would suggest going further and removing the requirement to provide a Java runtime at all. I think this makes sense for the following reasons:

1. As a developer I generally consume the implementations via maven or gradle build toolchains, both of which require a Java already and I'd much prefer to use the Java running maven and gradle to having a new one. 
2. As an implementer of an open source implementation I don't want the additional burden of distributing a JVM, and tracking security vulnerabilities for my user community. 
3. It isn't common for open source implementations to distribute a JVM which means they are cross platform compatible, requiring them to also distribute the JVM increases the download size and makes it platform specific.

starksm64:
The specification is overreaching in the prescription of the required runtime. It is the end user who chooses which JVM and environment they are running the EE container in, and therefore should be free to choose which underlying Java SE capabilities they which to include. Both sections 2.4.1 and 2.4.2 can be deleted in my opinion.

kwsutter:
This is old and no longer applies.  As we work through the rewording for support of Java SE 11 (ie. Issue #287), we will address the idea behind this PR as well.

PR #190, opened 2020-05-28
Platform overview updates for Jakarta EE 9

kwsutter:
Initial set of changes due to the Jakarta EE 9 release plan.  Still have some TODOs and Diagram updates, but this should be a good start...


kwsutter:
Still have some TODOs and Diagram updates required, but this is a good start.  Going to go ahead with the merge to make progress and we can do more cleanup later.

PR #192, opened 2020-05-31
update webprofile introduction for Jakarta EE 9

kazumura:
Signed-off-by: Kenji Kazumura <kzr@fujitsu.com>

### DIFF
--- a/specification/src/main/asciidoc/webprofile/Introduction.adoc
+++ b/specification/src/main/asciidoc/webprofile/Introduction.adoc
@@ -22,7 +22,7 @@ Besides managing HTTP interactions, most web
 applications have significant requirements in the areas of transaction
 management, security and persistence. Such requirements can be readily
 addressed by technologies that have been part of the Jakarta EE platform
-for quite some time, such as the Jakarta Enterprise Beans 4.x
+for quite some time, such as the Jakarta Enterprise Beans 3.x
 technology and the Jakarta Persistence, but that are rarely supported
 by “plain” servlet containers. By incorporating many of these APIs, the
 Web Profile aims at raising the bar for what should be considered a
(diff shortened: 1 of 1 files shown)


specification/src/main/asciidoc/webprofile/Introduction.adoc
```diff
@@ -93,7 +93,8 @@ from each grouping.
 For the first one, the Jakarta EE Platform
 specification mandates support for the “ _java:_ ” naming context in all
 profiles. Consequently, Web Profile products must support it. For a
-similar reason, all Web Profile 8 products must support the Java(TM) Platform, Standard Edition 8 API.
+similar reason, all Web Profile 9 products must support the Java(TM) Platform, Standard Edition 8 API,
```
smillidge, line 96: I don't think we said that products have to support JDK 8. We said apis have to be compiled for JDK 8.

specification/src/main/asciidoc/webprofile/Introduction.adoc
```diff
@@ -22,7 +22,7 @@ Besides managing HTTP interactions, most web
 applications have significant requirements in the areas of transaction
 management, security and persistence. Such requirements can be readily
 addressed by technologies that have been part of the Jakarta EE platform
-for quite some time, such as the Jakarta Enterprise Beans 3.x
+for quite some time, such as the Jakarta Enterprise Beans 4.x
```
kwsutter, line 25: This "3.x" reference was meant to indicate the alternate persistence framework defined by EJB 3 and JPA.  So, in this case, I would leave the "3.x" as-is and not update it to "4.x".

specification/src/main/asciidoc/webprofile/Introduction.adoc
kwsutter: Agree, @smillidge.  Maybe this can be simplified to saying "..all Web Profile 9 APIs must support the Java(TM) Platform, Standard Edition 8 API, and all Web Profile 9 products must run on Java(TM) Platform, Standard Edition 11 runtime.

PR #195, opened 2020-06-02
update transactions section per Jakarta EE 9 release plan

kwsutter:

Minor updates to fix up some inconsistencies in doc references.  Nothing directly related to Jakarta EE 9 release plan.

### DIFF
--- a/specification/src/main/asciidoc/platform/TransactionManagement.adoc
+++ b/specification/src/main/asciidoc/platform/TransactionManagement.adoc
@@ -28,7 +28,7 @@ for the Connectors specification, each component may also acquire one or more
 connections to access one or more transactional resource managers.
 
 For example, in
-_<<a475, Servlets/Server Pages Accessing Enterprise Beans>>_ , the call
+<<a475, Servlets/Server Pages Accessing Enterprise Beans>> , the call
 tree starts from a servlet or server pages accessing multiple enterprise beans,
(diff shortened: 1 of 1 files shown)


kwsutter:
> Hi @kwsutter, is the removal of the italics for consistency of rendering with other links? I guess so and don't have a problem with that. 

Correct.  I noticed that this section was not consistent with the use of italics for the doc references.  Since most of the other doc references in the Platform Spec are removing the italics, I decided to be consistent.

> Regarding the "<<" ">>" changes, what is the motivation for that? I think those links already look OK on https://jakarta.ee/specifications/platform/8/platform-spec-8.html#transaction-management so I am not sure? If there is a need to do it then I have no problem (even just to achieve consistency in the raw .adoc files) but I don't understand the motivation yet. Thanks!

@tomjenkinson , this was a change to help with future reading and editing of the asciidoc files.  In my editor, most of the asciidoc syntax is highlighted.  If the doc references `<<..>>` span more than one line, then the text is not properly highlighted.  This is probably a bug with my adoc-sensitive editor, but it's an easier fix to just put the doc references on a single line.  So, it's "syntactic sugar", but it helps with my constant reviewing and editing of these files.

Thanks for the review!

kwsutter:
Besides @tomjenkinson's technical review for Transactions, I also need an Approver with write access to the component to allow me to merge...  Thanks!

PR #204, opened 2020-06-04
Update Application Clients section per Jakarta EE 9 release plan

tjwatson:
- Add Jakarta 9 rows for annotation processing
 - Uppate link to application-client_9.xsd

Change-Id: I7e1342d482e7d2600097db4e45740de774800880
Signed-off-by: Thomas Watson <tjwatson@us.ibm.com>

### DIFF
--- a/specification/src/main/asciidoc/platform/ApplicationClients.adoc
+++ b/specification/src/main/asciidoc/platform/ApplicationClients.adoc
@@ -193,6 +193,14 @@ attribute.
 |No
 |Yes
 
+|application-client_9
+|Yes
+|No
+
+|application-client_9
+|No
+|Yes
+
 |none
 |N/A
 |Yes
@@ -303,7 +311,7 @@ file. Note that this name is case-sensitive.
 _<<a3404, Jakarta EE Application Client XML Schema Structure>>_ shows the structure of the Jakarta EE
 application-client XML Schema. The Jakarta EE application-client XML Schema
 is located at
-_http://xmlns.jcp.org/xml/ns/javaee/application-client_8.xsd_ .
+_https://jakarta.ee/xml/ns/jakartaee/application-client_9.xsd_.
 
 [[a3404]]
 .Jakarta EE Application Client XML Schema Structure


specification/src/main/asciidoc/platform/ApplicationClients.adoc
```diff
@@ -303,8 +311,8 @@ file. Note that this name is case-sensitive.
 _<<a3404, Jakarta EE Application Client XML Schema Structure>>_ shows the structure of the Jakarta EE
 application-client XML Schema. The Jakarta EE application-client XML Schema
 is located at
-_http://xmlns.jcp.org/xml/ns/javaee/application-client_8.xsd_ .
+_https://jakarta.ee/xml/ns/jakartaee/application-client_9.xsd_.
 
 [[a3404]]
 .Jakarta EE Application Client XML Schema Structure
-image::JavaEEapplication-client_schema_7.png[]
+image::JakartaEEapplication-client_schema_9.png[]
```
kwsutter, line 318: Shoot, @tjwatson, I don't know what to do with this file.  It's actually a representation of the schema at the java ee 7 version.  Since nothing changed in Java EE 8 or Jakarta EE 8, leaving it at "7" was expected.  Again, since nothing has changed in this release, I don't like updating it to "9".  And, since we're updating it to use the JakartaEE prefix, leaving it at "7" doesn't sound right either...  Maybe we should just leave the name of the file with "JavaEE", but leave the updated tagline at line 317 to say "Jakarta EE Application Client..."?  Am I overthinking this?

tjwatson: Maybe just make it JakartaEEapplication-client_schema with no version name?  I really have no idea how important any of this is.  The image name will not show up in the final doc.  If there is any question here then perhaps just leave it as is and keep the JavaEEapplication-client_schema_7.png link.  That image has to stay around so we can link to it from the previous versions section.

kwsutter: :-) That was my original thought -- to just drop the version.  But, like I said, I started to overthink this...  Here's my suggestion.  Either drop the version like you suggested, or just revert back to the "JavaEE" prefix and keep the "7" version.  Either way, you have to update the PR, so I'll let you make the decision.  :-)  Thanks!

tjwatson: I decided to revert back to the EE 7 image.  It has the advantage of not needing to make a copy of the image just to get a jakarta image name.  I think this is what we should do in other cases where the current Java EE image does not need modification.

PR #208, opened 2020-06-09
Insert an overall schedule to the Jakarta EE 9 page

kwsutter:

### DIFF
--- a/jakartaee9/JakartaEE9.md
+++ b/jakartaee9/JakartaEE9.md
@@ -30,28 +30,28 @@ These component release plans will utilize the existing Eclipse mechanism for de
 
 ## Jakarta EE 9 schedule
 
-Overall schedule (dates are tentative and subject to change).  The waves are based off of the content of the [Jakarta EE 9 Release Plan](JakartaEE9ReleasePlan).
+Overall schedule (dates are ***tentative*** and ***subject to change***).  The wave content is based off of the descriptions in the [Jakarta EE 9 Release Plan](JakartaEE9ReleasePlan).
 
 - Milestone 1 release (Java 8) - **June 23** announce
     - Development Complete - June 12
     - Testing Complete - June 19
-- Final Jakarta EE 9 release (Java 11) - **Sept 29** announce
-    - Wave 0 Specification PRs Complete - July 21
-    - Wave 0 Specification Ballots Complete - Aug 4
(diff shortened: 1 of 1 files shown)


jakartaee9/JakartaEE9.md
```diff
@@ -23,9 +28,34 @@ These component release plans will utilize the existing Eclipse mechanism for de
...
+## Jakarta EE 9 schedule
+
+Overall schedule (dates are tentative and subject to change).  The waves are based off of the content of the [Jakarta EE 9 Release Plan](JakartaEE9ReleasePlan).
+
+- Milestone 1 release (Java 8) - **June 23** announce
+    - Development Complete - June 12
+    - Testing Complete - June 19
+- Final Jakarta EE 9 release (Java 11) - **Sept 29** announce
+    - Wave 0 Specification PRs Complete - July 21
+    - Wave 0 Specification Ballots Complete - Aug 4
```
ivargrimstad, line 40: Arent wave 0 specs independent of the others and can be completed at any time? Could be sprinkled into any of the other wave batches? Just a suggestion

kwsutter: True, But that's hard to put into a schedule...  If you have ideas on how to represent the Wave 0 drops, I'm open to suggestions.  By leaving it first in the proposed schedule, then it allows them to come in later...

kwsutter:
Based on the discussion with the Steering Committee yesterday, I'm going to adjust the final dates to line up with the scheduled JakartaOne Livestream on Sept 16.  Since these dates are a risk regardless, adjusting the final date by 6 days is not going to affect the outcome.  We need Jakarta EE 9 to be final on or before the Livestream event.

jakartaee9/JakartaEE9.md
kwsutter: I did end up changing this slightly.  When I adjusted the final date to be consistent with the Livestream event, then Wave 0 just became too early in July.  So, I just indicated that Wave 0 has to be done sometime between now and the end of Wave 6.

kwsutter:
Since the dates were slight adjusted based on the Steering committee discussion, I'm going to ask for another re-review from Ivar and Steve.  I also adjusted the bullets for wave 0 to show that these can be done at any time.  Thanks.

PR #225, opened 2020-07-20
add 'Ballots Start' to Schedule

kwsutter:

Just clarifying that when the PRs are Complete, then the Ballots should also Start.


edbratt:
I thought ballots started on Tuesdays -- so, complete the day previous and the ballots start on the next Tuesday. Did I miss a relaxation of this point?

PR #248, opened 2020-10-01
Update JTS 1.0 link

kwsutter:

### DIFF
--- a/specification/src/main/asciidoc/platform/RelatedDocuments.adoc
+++ b/specification/src/main/asciidoc/platform/RelatedDocuments.adoc
@@ -30,7 +30,7 @@ _Jakarta™ Messaging Specification, Version 3.0_. Available at: _https://jakart
 
 _Jakarta™ Transaction Specification, Version 2.0_. Available at: _https://jakarta.ee/specifications/transactions/2.0_
 
-_Java™ Transaction Service, Version 1.0_ (JTS specification)_. Available at: _https://www.oracle.com/technetwork/java/javaee/jts-spec095-1508547.pdf_
+_Java™ Transaction Service, Version 1.0_ (JTS specification)_. Available at: _http://download.oracle.com/otn-pub/jcp/7309-jts-1.0-spec-oth-JSpec/jts1_0-spec.pdf_
 
 _Jakarta™ Mail Specification, Version 2.0_. Available at: _https://jakarta.ee/specifications/mail/2.0_
 


tomjenkinson (commented):
I am not sure actually, the Jakarta Transactions specification links to 0.95 so perhaps we should keep that and (maybe) remove the ", Version 1.0" part of this link? We could just align it with https://github.com/eclipse-ee4j/jta-api/blame/master/spec/src/main/asciidoc/Transactions.adoc#L1390-L1391 for consistency? Do you agree?

kwsutter:
> I would note that JTS was only a draft specification AFAIK so alternatively, if you want to keep1.0 I think we would probably want to consider changing it to be something like "1.0 (DRAFT)"

I like this compromise, @tomjenkinson.  Thanks for the idea.  If you would then update the Transactions Spec in a similar vein, then we would be in sync.

specification/src/main/asciidoc/platform/RelatedDocuments.adoc
```diff
@@ -30,7 +30,7 @@ _Jakarta™ Messaging Specification, Version 3.0_. Available at: _https://jakart
 
 _Jakarta™ Transaction Specification, Version 2.0_. Available at: _https://jakarta.ee/specifications/transactions/2.0_
 
-_Java™ Transaction Service, Version 1.0_ (JTS specification)_. Available at: _https://www.oracle.com/technetwork/java/javaee/jts-spec095-1508547.pdf_
+_Java™ Transaction Service, Version 1.0 (Draft)_ (JTS specification)_. Available at: _http://download.oracle.com/otn-pub/jcp/7309-jts-1.0-spec-oth-JSpec/jts1_0-spec.pdf_
```
tomjenkinson, line 33: I think that this issue was likely already present but I think the "_" after the (Draft) should be removed.

To match Jakarta Transactions (https://github.com/eclipse-ee4j/jta-api/pull/177/files) then it would be something like:

`_Java™ Transaction Service (JTS) Specification, Version 1.0 (Draft)_. Available at: _http://download.oracle.com/otn-pub/jcp/7309-jts-1.0-spec-oth-JSpec/jts1_0-spec.pdf_`

There are some formatting differences and the reference in your file has a TM after Java but in Transactions we have this in a separate file: https://github.com/eclipse-ee4j/jta-api/blame/master/spec/src/main/asciidoc/license-efsl.adoc#L17-L18. I think generally though the change here and in the Jakarta Transactions project would then be aligned.

Perhaps we need the specification change merged first: https://github.com/eclipse-ee4j/jta-api/pull/177

tomjenkinson: The Jakarta Transactions PR is merged. So if you think it's appropriate you could refer to what we are doing: https://github.com/eclipse-ee4j/jta-api/blame/master/spec/src/main/asciidoc/Transactions.adoc#L1390-L1391

kwsutter: All of the references in the Platform are italicized (with the underscores), so I'm going to leave them as-is.  I don't think there's a requirement to be consistent across all specs with that convention.  I will move the (JTS) abbreviation earlier in my commit to be consistent.  Otherwise, I think we're good.  Thanks!

kwsutter: As I was making my edit, I noticed what you were referring to with the underscore after the (Draft).  Thanks for catching that!

ivargrimstad (approved):

specification/src/main/asciidoc/platform/RelatedDocuments.adoc
```diff
@@ -30,7 +30,7 @@ _Jakarta™ Messaging Specification, Version 3.0_. Available at: _https://jakart
 
 _Jakarta™ Transaction Specification, Version 2.0_. Available at: _https://jakarta.ee/specifications/transactions/2.0_
 
-_Java™ Transaction Service, Version 1.0_ (JTS specification)_. Available at: _https://www.oracle.com/technetwork/java/javaee/jts-spec095-1508547.pdf_
+_Java™ Transaction Service (JTS) Specification, Version 1.0 (Draft)_. Available at: _http://download.oracle.com/otn-pub/jcp/7309-jts-1.0-spec-oth-JSpec/jts1_0-spec.pdf_
```
tomjenkinson, line 33: I think we should change the "http://" to "https://" I am going to raise a PR for the same on Jakarta Transactions

tomjenkinson (changes requested):

specification/src/main/asciidoc/platform/RelatedDocuments.adoc
kwsutter: Doh!  I saw that when I was making my other edits, but I must have missed pushing the change.  Sorry and thanks!

ivargrimstad (approved):

tomjenkinson (approved):
Thanks!

kwsutter:
I'm going to go ahead with this merge based on Ivar's approval and resolving Tom's "last" comment.  If there are more changes necessary, then we can do that via another PR.  Thanks!

PR #249, opened 2020-10-01
Complete section on RMI-IIOP and add some history

dblevins:
This is perhaps a bit redundant, but given the history on this feature and reasonable confusion I think there's some benefit to having the requirements restated a couple times.  The line between portability and interoperability needs to be 100% clear.  RMI-IIOP is no longer relevant for interoperability, but still very relevant to portability.

RMI-IIOP programming restrictions remain intact and allow Jakarta EE implementations freedom to chose CORBA or any other protocol.

A concrete example is a Jakarta EE implementation could use HTTP and application/json as their underlying protocol, however, they would need to support RMI-IIOP semantics:  they must support `java.io.Serializable`, `java.io.Externalizable` and call all associated callbacks such as  `readObject` and `replaceObject` detailed by those interfaces; they must not serialize/deserialize classes that do not implement Serializable or Externalizable;  they must retain the references in the serialized object graph entirely upon deserialization.

Maintaining references within an object graph is something text protocols typically cannot do and is likely to be a source of portability issues should more protocols enter our ecosystem.  As an example early SOAP implementations used SOAP Encoded semantics and could handle complex reference situations such as the same object being in a map 4 times and deserialized cleanly as one map containing four references to the same object.  Later SOAP implementations favored "doc literal wrapped", which would handle the same situation by completely serializing the object four times resulting in a deserialized map that had four copies of the object resulting in a deserialized object graph that is fundamentally different from the serialized object graph.  This later behavior would not be OK from the eyes of RMI or IIOP (referred to as RMI-IIOP for shorthand).

We currently do not have significant tests for these guarantees.  These tests are not there because they were part of the RMI-IIOP interoperability tests, which are removed as they are unfortunately also CORBA specific.  We would need a new suite that is exclusively focused on portability, not interoperability.  Should we find ourselves in a situation where new protocols find their way into the next generation of Jakarta EE implementations and people notice portability in serialization/deserialization semantics (a deserialized object graph is not quite the same when porting from one server to another), we will need to expand our tests.

### DIFF
--- a/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
+++ b/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
@@ -390,9 +390,6 @@ The following Jakarta EE Technologies were removed from the Jakarta EE Platform.
 
 |Distributed Interoperability (EJB 3.2 Core Specification, Chapter 10)
 |Removed in Jakarta EE 9
-
-|CORBA, Java IDL, RMI-IIOP
-|Removed in Jakarta EE 9
 |===
 
 [[a2339]]
--- a/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
+++ b/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
@@ -1001,7 +1001,7 @@ application component types can be clients of RMI objects.
 
 The RMI-IIOP subsystem is composed of APIs that allow for the
 use of RMI-style programming that is independent of the underlying
-protocol.  Implementations these APIs may support the Java SE native RMI
+protocol.  Implementations of these APIs may support the Java SE native RMI
 protocol (JRMP), the CORBA IIOP protocol or any custom protocol that is
 compatible with the RMI programming restrictions (see the RMI-IIOP spec
 for details).
@@ -1010,7 +1010,7 @@ NOTE: The requirements in this section only apply to Jakarta EE products that
 include an Enterprise Beans container with support for remote interfaces.
 
 Jakarta EE applications use the RMI-IIOP APIs when accessing
-remote Enterprise JavaBeans components, as described in the Jakarta Enterprise
+remote Enterprise Beans components, as described in the Jakarta Enterprise
 Beans 4.0 specification.  This allows Enterprise Beans and their clients to be
 protocol independent and portable to Jakarta EE implementations that may use
 CORBA/IIOP, RMI, or any other custom protocol.  However, there is currently no
@@ -1024,7 +1024,7 @@ programming restrictions necessary to allow application portability.  Requiremen
 for implementations to use CORBA/IIOP itself were added to J2EE 1.3 and EJB 2.0 and
 expanded the term RMI-IIOP beyond application portability and into distributed
 interoperability. These distributed interoperability requirements
-were marked Proposed Optional in Java EE 1.8 .  As of Jakarta EE 9 and Enterprise
+were marked Proposed Optional in Java EE 8 .  As of Jakarta EE 9 and Enterprise
 Beans 4.0 distributed interoperability requirements are fully removed and the term
 RMI-IIOP reverts back to a term that guarantees only application portability, not
 distributed interoperability.  Remaining use of the narrow method of
--- a/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
+++ b/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
 
 ===== RMI-JRMP
 
-*TODO Note:* What to do with this section? Leave as-is, mark Optional, remove?
-
 JRMP is the Java technology-specific Remote
 Method Invocation (RMI) protocol. The Jakarta EE security restrictions
 typically prevent all application component types except application
 clients from creating and exporting an RMI object, but all Jakarta EE
 application component types can be clients of RMI objects.
 
-===== RMI-IIOP (Proposed Optional)
+===== RMI-IIOP
 
-*TODO Note:* Since distributed interop is being removed from the EJB spec, do we remove
-this section completely?
+The RMI-IIOP subsystem is composed of APIs that allow for the
+use of RMI-style programming that is independent of the underlying
+protocol.  Implementations these APIs may support the Java SE native RMI
+protocol (JRMP), the CORBA IIOP protocol or any custom protocol that is
+compatible with the RMI programming restrictions (see the RMI-IIOP spec
+for details).
 
-The requirements in this section only apply
-to Jakarta EE products that include an Enterprise Beans container and support
-interoperability using RMI-IIOP.
+NOTE: The requirements in this section only apply to Jakarta EE products that
+include an Enterprise Beans container with support for remote interfaces.
 
-RMI-IIOP allows objects defined using RMI
-style interfaces to be accessed using the IIOP protocol. It must be
-possible to make any remote _enterprise bean accessible via_ RMI-IIOP.
-Some Jakarta  EE products will simply make all remote enterprise beans
-always (and only) accessible via RMI-IIOP; other products might control
-this via an administrative or deployment action. These and other
-approaches are allowed, provided that any remote enterprise bean (or by
-extension, all remote enterprise beans) can be made accessible using
-RMI-IIOP.
+Jakarta EE applications are required to use the RMI-IIOP APIs (specifically the
+narrow method of javax.rmi.PortableRemoteObject) when accessing
+remote Enterprise JavaBeans components, as described in the Jakarta Enterprise
+Beans 4.0 specification.  This allows Enterprise Beans and their clients to be
+protocol independent and portable to Jakarta EE implementations that may use
+CORBA or RMI as their underlying protocol.  However, there is currently no
+requirement that Enterprise Beans be accessible using the IIOP protocol itself
+or that Jakarta EE implementations support any form of distributed interoperability
+with other Jakarta EE implementations.
 
-Components accessing remote _enterprise
-beans_ may need to use the _narrow_ method of the
-_javax.rmi.PortableRemoteObject_ class, under circumstances described in
-the Jakarta Enterprise Beans specification. Because remote enterprise beans may be deployed
-using other RMI protocols, portable applications must not depend on the
-characteristics of RMI-IIOP objects (for example, the use of the _Stub_
-and _Tie_ base classes) beyond what is specified in the Jakarta Enterprise Beans
-specification.
+The use of the term RMI-IIOP has existed since J2EE 1.2 and EJB 1.1, predates
+any requirements to support CORBA itself and was used to detail application
+programming restrictions necessary to allow application portability.  Requirements
+for implementations to use CORBA IIOP itself were added to J2EE 1.3 and EJB 2.0 and
+expanded the term RMI-IIOP beyond application portability and into distributed
+interoperability. These distributed interoperability requirements
+were marked Proposed Optional in Java EE 1.8 .  As of Jakarta EE 9 and Enterprise
+Beans 4.0 distributed interoperability requirements are fully removed and the term
+RMI-IIOP reverts back to a term that guarantees only application portability, not
+distributed interoperability.
 
-The Jakarta EE security restrictions typically
-prevent all application component types, except application clients,
-from creating and exporting an RMI-IIOP object. All Jakarta EE application
-component types can be clients of RMI-IIOP objects. Jakarta EE applications
-should also use JNDI to lookup non-Enterprise Beans RMI-IIOP objects. The JNDI names
-used for such non-Enterprise Beans RMI-IIOP objects should be configured at
-deployment time using the standard environment entries mechanism (see
-<<a607, JNDI Naming Context>>).
-The application should fetch a name from JNDI using an environment
-entry, and use the name to lookup the RMI-IIOP object. Typically such
-names will be configured to be names in the COSNaming name service.
-
-This specification does not provide a
-portable way for applications to bind objects to names in a name
-service. Some products may support use of JNDI and COSNaming for binding
-objects, but this is not required. Portable Jakarta EE application clients
-can create non-Enterprise Beans RMI-IIOP server objects for use as callback objects,
-or to pass in calls to other RMI-IIOP objects.
-
-Note that while RMI-IIOP doesn’t specify how
-to propagate the current security context or transaction context, the
-Jakarta Enterprise Beans interoperability specification does define such context propagation.
-This specification only requires that the propagation of context
-information as defined in the Jakarta Enterprise Beans specification be supported in the use
-of RMI-IIOP to access enterprise beans. The propagation of context
-information is not required in the uses of RMI-IIOP to access objects
-other than enterprise beans.
-
-The RMI-IIOP specification describes how
-portable Stub and _Tie_ classes can be created. To be portable to all
-implementations that use a CORBA Portable Object Adapter (POA), the
-_Tie_ classes must extend the _org.omg.PortableServer.Servant_ class.
-This is typically done by using the _-poa_ option to the _rmic_ command.
-A Jakarta EE product must provide support for these portable _Stub_ and
-_Tie_ classes, typically using the required CORBA POA. However, for
-portability to systems that do not use a POA to implement RMI-IIOP,
-applications should not depend on the fact that the _Tie_ extends the
-_Servant_ class. A Jakarta EE application that defines or uses RMI-IIOP
-objects other than enterprise beans must include such portable _Stub_
-and _Tie_ classes in the application package. _Stub_ and _Tie_ objects
-for enterprise beans, however, must not be included with the
-application: they will be generated, if needed, by the Jakarta EE product
-at deployment time or at run time.
-
-RMI-IIOP is defined by chapters 5, 6, 13, 15,
-and section 10.6.2 of the CORBA 2.3.1 specification, available at
-_http://www.omg.org/cgi-bin/doc?formal/99-10-07_ , and by the Java™
-Language To IDL Mapping Specification, available at
-_http://www.omg.org/cgi-bin/doc?ptc/2000-01-06_ .
+Jakarta EE implementations may use CORBA IIOP as their underlying protocol, however,
+such support is implementation-specific and no longer a guarantee of the Jakarta
+EE platform.
 
 ===== JNDI
 
--- a/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
+++ b/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
@@ -89,12 +89,6 @@ This is a stronger statement than making a technology "optional", since a "remov
 technology will no longer be maintained for future versions of the Platform.
 <<a2333, Removed Jakarta Technologies>> documents these removed technologies.
 
-*Table TODO* There doesn't seem to be any rhyme or reason to the ordering of the contents
-of this table.  Maybe it was done in the order of the specs getting added to the platform?
-Maybe we should just alphabetize the technologies? Reminder -- If the rows in this table
-get re-organized, then the Requirements sections at the end of this API section also need to
-be similarly re-organized.
-
 [[a2159]]
 [cols=5, options=header]
 .Jakarta EE Technologies
@@ -1018,12 +1012,11 @@ for details).
 NOTE: The requirements in this section only apply to Jakarta EE products that
 include an Enterprise Beans container with support for remote interfaces.
 
-Jakarta EE applications are required to use the RMI-IIOP APIs (specifically the
-narrow method of javax.rmi.PortableRemoteObject) when accessing
+Jakarta EE applications use the RMI-IIOP APIs when accessing
 remote Enterprise JavaBeans components, as described in the Jakarta Enterprise
 Beans 4.0 specification.  This allows Enterprise Beans and their clients to be
 protocol independent and portable to Jakarta EE implementations that may use
-CORBA or RMI as their underlying protocol.  However, there is currently no
+CORBA/IIOP, RMI, or any other custom protocol.  However, there is currently no
 requirement that Enterprise Beans be accessible using the IIOP protocol itself
 or that Jakarta EE implementations support any form of distributed interoperability
 with other Jakarta EE implementations.
@@ -1031,15 +1024,16 @@ with other Jakarta EE implementations.
 The use of the term RMI-IIOP has existed since J2EE 1.2 and EJB 1.1, predates
 any requirements to support CORBA itself and was used to detail application
 programming restrictions necessary to allow application portability.  Requirements
-for implementations to use CORBA IIOP itself were added to J2EE 1.3 and EJB 2.0 and
+for implementations to use CORBA/IIOP itself were added to J2EE 1.3 and EJB 2.0 and
 expanded the term RMI-IIOP beyond application portability and into distributed
 interoperability. These distributed interoperability requirements
 were marked Proposed Optional in Java EE 1.8 .  As of Jakarta EE 9 and Enterprise
 Beans 4.0 distributed interoperability requirements are fully removed and the term
 RMI-IIOP reverts back to a term that guarantees only application portability, not
-distributed interoperability.
+distributed interoperability.  Remaining use of the narrow method of
+`javax.rmi.PortableRemoteObject` is slated for removal in a future release.
 
-Jakarta EE implementations may use CORBA IIOP as their underlying protocol, however,
+Jakarta EE implementations may use CORBA/IIOP as their underlying protocol, however,
 such support is implementation-specific and no longer a guarantee of the Jakarta
 EE platform.
 
@@ -1185,37 +1179,16 @@ applications.
 
 === Enterprise Beans 4.0 Requirements
 
-*TODO Note:*  This section still needs some work due to the revamping of the EJB 4.0
-spec for Jakarta EE 9...
-
 This specification requires that a  Jakarta EE
 product provide support for _enterprise beans_ as specified in the Jakarta Enterprise
 Beans specification. The Jakarta Enterprise Beans specification is available at
 _https://jakarta.ee/specifications/enterprise-beans_ .
 
-This specification does not impose any
-additional requirements at this time. Note that the Jakarta Enterprise Beans
-specification includes the specification of the Jakarta Enterprise Beans
-interoperability protocol based on RMI-IIOP. Support for the Jakarta Enterprise Beans
-interoperability protocol is Proposed Optional in Jakarta EE 8. All containers that
-support Jakarta Enterprise Beans clients must be capable of using the
-Jakarta Enterprise Beans interoperability protocol to invoke enterprise
-beans. All Jakarta Enterprise Beans containers must support the invocation of enterprise
-beans using the Jakarta Enterprise Beans interoperability protocol. A Jakarta EE
-product may also support other protocols for the invocation of enterprise beans.
-
-A Jakarta EE product may support multiple object
-systems (for example, RMI-IIOP and RMI-JRMP). It may not always be
-possible to pass object references from one object system to objects in
-another object system. However, when an enterprise bean is using the
-RMI-IIOP protocol, it must be possible to pass object references for
-RMI-IIOP or Java IDL objects as arguments to methods on such an
-enterprise bean, and to return such object references as return values
-of a method on such an enterprise bean. In addition, it must be possible
-to pass a reference to an RMI-IIOP-based enterprise bean’s Home or
-Remote interface to a method on an RMI-IIOP or Java IDL object, or to
-return such an enterprise bean object reference as a return value from
-such an RMI-IIOP or Java IDL object.
+A Jakarta EE product may support multiple object systems (for example,
+RMI-IIOP, RMI-JRMP, gRPC, protobuf, Thrift).  There is no explicit
+requirement that a Jakarta EE product support any specific protocol,
+such as CORBA/IIOP, or provide distributed interoperability between
+products.
 
 In a Jakarta EE product that includes both an
 enterprise beans container and a web container, both containers are required to


specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+protocol (JRMP), the CORBA IIOP protocol or any custom protocol that is
+compatible with the RMI programming restrictions (see the RMI-IIOP spec
+for details).
+
+NOTE: The requirements in this section only apply to Jakarta EE products that
+include an Enterprise Beans container with support for remote interfaces.
+
+Jakarta EE applications are required to use the RMI-IIOP APIs (specifically the
+narrow method of javax.rmi.PortableRemoteObject) when accessing
+remote Enterprise JavaBeans components, as described in the Jakarta Enterprise
```
hussainnm, line 1023: Should this be Enterprise Beans?

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
-RMI-IIOP is defined by chapters 5, 6, 13, 15,
-and section 10.6.2 of the CORBA 2.3.1 specification, available at
-_http://www.omg.org/cgi-bin/doc?formal/99-10-07_ , and by the Java™
-Language To IDL Mapping Specification, available at
-_http://www.omg.org/cgi-bin/doc?ptc/2000-01-06_ .
+===== RMI-IIOP
+
+The RMI-IIOP subsystem is composed of APIs that allow for the
+use of RMI-style programming that is independent of the underlying
+protocol.  Implementations these APIs may support the Java SE native RMI
```
anthonyvdotbe, line 1013:
```suggestion
protocol.  Implementations of these APIs may support the Java SE native RMI
```

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+
+Jakarta EE applications are required to use the RMI-IIOP APIs (specifically the
+narrow method of javax.rmi.PortableRemoteObject) when accessing
+remote Enterprise JavaBeans components, as described in the Jakarta Enterprise
+Beans 4.0 specification.  This allows Enterprise Beans and their clients to be
+protocol independent and portable to Jakarta EE implementations that may use
+CORBA or RMI as their underlying protocol.  However, there is currently no
+requirement that Enterprise Beans be accessible using the IIOP protocol itself
+or that Jakarta EE implementations support any form of distributed interoperability
+with other Jakarta EE implementations.
```
kwsutter, line 1029: This paragraph is confusing to me.  The first sentence says that Jakarta EE applications are **required** to use RMI-IIOP for remote EJBs.  But, then you conclude that there is no requirement for EJBs to be accessible via the IIOP protocal, nor that app servers need to support distributed interop.  Maybe it's just me, but I just don't get this distinction.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+The use of the term RMI-IIOP has existed since J2EE 1.2 and EJB 1.1, predates
+any requirements to support CORBA itself and was used to detail application
+programming restrictions necessary to allow application portability.  Requirements
+for implementations to use CORBA IIOP itself were added to J2EE 1.3 and EJB 2.0 and
+expanded the term RMI-IIOP beyond application portability and into distributed
+interoperability. These distributed interoperability requirements
+were marked Proposed Optional in Java EE 1.8 .  As of Jakarta EE 9 and Enterprise
+Beans 4.0 distributed interoperability requirements are fully removed and the term
+RMI-IIOP reverts back to a term that guarantees only application portability, not
+distributed interoperability.
```
kwsutter, line 1040: Overall, I really like this paragraph.  This helps with the history and understanding of how we got where we are.  But, I'm still confused on the app portability vs distributed interop definitions and how they apply to application servers implementing Jakarta EE 9, and how they apply to applications expected to run on Jakarta EE 9 runtimes.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
-names will be configured to be names in the COSNaming name service.
-
-This specification does not provide a
-portable way for applications to bind objects to names in a name
-service. Some products may support use of JNDI and COSNaming for binding
-objects, but this is not required. Portable Jakarta EE application clients
-can create non-Enterprise Beans RMI-IIOP server objects for use as callback objects,
-or to pass in calls to other RMI-IIOP objects.
-
-Note that while RMI-IIOP doesn’t specify how
```
kwsutter, line 1058: These security aspects of RMI-IIOP are not referenced in the replacement text for this section.  Is this required?  Or, is it sufficiently covered in the EJB spec (or elsewhere) and this was just redundant?

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
-portability to systems that do not use a POA to implement RMI-IIOP,
-applications should not depend on the fact that the _Tie_ extends the
-_Servant_ class. A Jakarta EE application that defines or uses RMI-IIOP
-objects other than enterprise beans must include such portable _Stub_
-and _Tie_ classes in the application package. _Stub_ and _Tie_ objects
-for enterprise beans, however, must not be included with the
-application: they will be generated, if needed, by the Jakarta EE product
-at deployment time or at run time.
-
-RMI-IIOP is defined by chapters 5, 6, 13, 15,
```
kwsutter, line 1083: Do we still need these references in the replacement write-up?  Seems like they still may be required, especially if "portable applications" or "remote EJBs" will need to understand RMI-IIOP...

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+expanded the term RMI-IIOP beyond application portability and into distributed
+interoperability. These distributed interoperability requirements
+were marked Proposed Optional in Java EE 1.8 .  As of Jakarta EE 9 and Enterprise
+Beans 4.0 distributed interoperability requirements are fully removed and the term
+RMI-IIOP reverts back to a term that guarantees only application portability, not
+distributed interoperability.
+
+Jakarta EE implementations may use CORBA IIOP as their underlying protocol, however,
+such support is implementation-specific and no longer a guarantee of the Jakarta
+EE platform.
```
kwsutter, line 1045: Other sections need attention relating to Interop... Will they be separate PRs?  Or, do you plan to enhance this PR?  

Section 2.8 Interoperability (text and diagram)
Section 2.13.3 Network Protocols
Section 6.1.4 Removed Jakarta Technologies (needs clarification on CORBA, Java IDL, RMI-IIOP)
Section 6.3 Enterprise Beans 4.0 Requirements
Section 7.2 Interoperability Protocols
Section 7.2.3 Java Technology Protocols
Section 9.6 Optional Features for Jakarta EE Profiles

kwsutter (changes requested):
@dblevins Thanks for taking a shot at this effort!  This Jakarta EE 9 change has quite the ripple effect across the Platform and EJB specifications.  You are definitely more in tune with these protocols than I am.  But, I'm still quite confused on how we present this material to our readers and how it affects the app servers as well as the applications.  Thanks!

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+or that Jakarta EE implementations support any form of distributed interoperability
+with other Jakarta EE implementations.
+
+The use of the term RMI-IIOP has existed since J2EE 1.2 and EJB 1.1, predates
+any requirements to support CORBA itself and was used to detail application
+programming restrictions necessary to allow application portability.  Requirements
+for implementations to use CORBA IIOP itself were added to J2EE 1.3 and EJB 2.0 and
+expanded the term RMI-IIOP beyond application portability and into distributed
+interoperability. These distributed interoperability requirements
+were marked Proposed Optional in Java EE 1.8 .  As of Jakarta EE 9 and Enterprise
```
Emily-Jiang, line 1037: Java EE 1.8 should be Java EE8.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
Emily-Jiang: @kwsutter I think @dblevins meant Jakarta EE application uses the API without mandating the protocol, as the API impl could be CORBA or RMI.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
-for enterprise beans, however, must not be included with the
-application: they will be generated, if needed, by the Jakarta EE product
-at deployment time or at run time.
-
-RMI-IIOP is defined by chapters 5, 6, 13, 15,
-and section 10.6.2 of the CORBA 2.3.1 specification, available at
-_http://www.omg.org/cgi-bin/doc?formal/99-10-07_ , and by the Java™
-Language To IDL Mapping Specification, available at
-_http://www.omg.org/cgi-bin/doc?ptc/2000-01-06_ .
+===== RMI-IIOP
```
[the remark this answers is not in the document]
Emily-Jiang: I think this section should clearly state that RMI-IIOP APIs is supported but not interop. I think this was the confusion from @kwsutter , if I understand correctly.

scottmarlow:
> We currently do not have significant tests for these guarantees. These tests are not there because they were part of the RMI-IIOP interoperability tests, which are removed as they are unfortunately also CORBA specific. We would need a new suite that is exclusively focused on portability, not interoperability. Should we find ourselves in a situation where new protocols find their way into the next generation of Jakarta EE implementations and people notice portability in serialization/deserialization semantics (a deserialized object graph is not quite the same when porting from one server to another), we will need to expand our tests.

@dblevins we also have other RMI-IIOP/Corba tests that we are waiting for input on removing.  As mentioned on https://www.eclipse.org/lists/ejb-dev/msg00136.html, instead of making the RMI-IIOP + Corba tests optional, we are removing them to be consistent with how we handled other pruned/removed technologies.  I'm bringing attention to this conversation in this thread so others can follow that conversation as it relates to whatever we decide here.  Feedback on this other linked conversation is very much welcome in the ejb-dev mailing list thread (not here, unless it becomes important to this pull request).  :-)

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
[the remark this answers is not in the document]
scottmarlow: javax.rmi.PortableRemoteObject is still required for EJBs, however the Corba IIOP protocol is not required or optional for EE 9.  

IMO, `RMI-IIOP` is now badly named for our purposes.  It came out after Java RMI + the JRMP protocol and was meant to add to what `RMI` had already defined.  Perhaps a better name would of been `RMI-IIOP_or_NO_IIOP` (I'm kidding about the better name suggestion :-).

Pasting the description from https://docs.oracle.com/javase/8/docs/api/javax/rmi/package-summary.html which refers to javax.rmi.PortableRemoteObject:
> Package javax.rmi Description
Contains user APIs for RMI-IIOP. These APIs are provided for use by RMI-IIOP applications, and provide equivalent semantics when running over either IIOP or JRMP. See also the javax.rmi.CORBA package.


In case it gives us some ideas of what to call this section, `RMI-IIOP` as [defined from wikipedia](https://en.wikipedia.org/wiki/RMI-IIOP):
> RMI-IIOP (read as "RMI over IIOP") denotes the Java Remote Method Invocation (RMI) interface over the Internet Inter-Orb Protocol (IIOP), which delivers Common Object Request Broker Architecture (CORBA) distributed computing capabilities to the Java platform. It was initially based on two specifications: the Java Language Mapping to OMG IDL, and CORBA/IIOP 2.3.1.[1]
>
> With features inherited from CORBA, software components that work together can be written in multiple computer languages and run on multiple computers. In other words, it supports multiple platforms and can make remote procedure calls to execute, subroutines on another computer as defined by RMI. 
```

I will add some possible suggestions for changing the title from RMI-IIOP to something else.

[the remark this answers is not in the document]
Comment (scottmarlow) [reply specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc:1009]: Title change:
===== RMI
Or
===== Remote invocations with Jakarta Enterprise Beans
Or 
===== Remoting with Jakarta Enterprise Beans

File: specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc [hunk 10]
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+
+The RMI-IIOP subsystem is composed of APIs that allow for the
```
scottmarlow, line 1011: I'm not sure of what `The RMI-IIOP subsystem` is.  Could this be changed to `RMI-IIOP is composed of APIs that allow for the`?

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
scottmarlow: Do we need to state that is no requirement for..., could we instead state:
```
Beans 4.0 specification. This allows Enterprise Beans and their clients to be protocol independent and portable to Jakarta EE implementations using implementation specific protocols.
```

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
scottmarlow: Could we end the sentence after `removed`?  

So, something like:
`As of Jakarta EE 9 and Enterprise Beans 4.0 distributed interoperability requirements are fully removed.'

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+expanded the term RMI-IIOP beyond application portability and into distributed
+interoperability. These distributed interoperability requirements
+were marked Proposed Optional in Java EE 1.8 .  As of Jakarta EE 9 and Enterprise
+Beans 4.0 distributed interoperability requirements are fully removed and the term
+RMI-IIOP reverts back to a term that guarantees only application portability, not
+distributed interoperability.
+
+Jakarta EE implementations may use CORBA IIOP as their underlying protocol, however,
+such support is implementation-specific and no longer a guarantee of the Jakarta
+EE platform.
```
scottmarlow, line 1044: Do we need to include what EE 9 no longer guarantees here?  

Perhaps changing to:
> Jakarta EE implementations may use CORBA IIOP as their implementation-specific (underlying) protocol.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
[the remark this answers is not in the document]
kwsutter: @scottmarlow Thanks!  As I continue to read about this, I tend to agree with you that we will need to continue to support RMI, but the RMI-IIOP combination can be removed.  If that's the case, then that would help clarify the situation.  @dblevins?  Are we off base?

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
kwsutter: As long as we clearly differentiate between distributed interop and application portability (hopefully earlier in this section), then we could end the sentence after "removed".

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
kwsutter: Nice catch, Emily.  I would just be sure to put a space before the 8...  "Java EE 8".

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
scottmarlow, line 1009: Rewriting my original comment on this section, so deleted my previous comment rather than updating, not sure if this is clearer or not:

As per https://openjdk.java.net/jeps/320, `javax.rmi.PortableRemoteObject` is removed from Java SE 11 (which is indeed part of Corba/RMI-IIOP).

Whether we drop or rename or keep the RMI-IIOP section, could we refer to the EJB spec dependency on `javax.rmi.PortableRemoteObject` as `legacy` support?  Using the `optional` word to describe the EJB dependency on `javax.rmi.PortableRemoteObject` implies too much (e.g. having javax.rmi.PortableRemoteObject using TCK tests that only run on Java SE 8).

starksm64:
In my opinion none of the RMI/IIOP APIs should remain since there are no useful required tests for them and really no point without a standard interop protocol. If a future version does introduce a standard remoting protocol, we can deal with the requirements at that point.

kwsutter:
> In my opinion none of the RMI/IIOP APIs should remain since there are no useful required tests for them and really no point without a standard interop protocol. If a future version does introduce a standard remoting protocol, we can deal with the requirements at that point.

This approach would definitely simplify things for Jakarta EE 9.  I just don't know if this approach would satisfy the requirements for remote EJBs.  @dblevins?

dblevins:
The following is not a direct response to anyone, but is intended to provide enough context that can hopeully answer some of the questsions asked.  If any of your questions are answered by this, please indicate that and mark them resolved.  If you do read this and it does not answer your question, please update your question and I'll follow up.

# EJB 1.0

##  Heavy RMI Focus

There are 113 references to "RMI". These are throughout the spec in the description of any user-accessible method from a Home or Remote.  Note there were no local interfaces yet and all EJB interfaces were remote.  Since that made any method invoked on an EJB serialized and serialization is tricky, there needed to be explicit rules on what is safe for the wire.  The spec is filled with language that describes what developer requirements such as, "The methods arguments must be legal types for Java RMI."  Also note there are no Entity beans yet and the specification is only 181 pages, so 113 RMI references is basically the entire spec.  At this time

## Light but clear CORBA Goal

There are 23 references to "CORBA" and 7 references to IIOP.  There was a clear statement in the Introduction section of the spec that said, "The Enterprise JavaBeans architecture will be compatible with CORBA."  Here "compatible with" does not mean the EJB developer can expect COBRA to be the protocol the server uses, but that the server *may* use COBRA and it's the EJB specifications intent to allow that and further define mappings so people know how this should work.  There is a section titled "Interoperability via network protocol", however this section is really a description of the benefits of COBRA.  The section "Standard CORBA mapping" contains the majority of requirements if IIOP is used.


## Open to Other Protocols

There is a section titled "Support for other protocols" and explicitly states the vendor may use any protocol it likes.  It even gives an example of using HTTP to invoke EJBs or exposing EJBs to "OLE Automation clients, such as Visual Basic."  It also says the mapping to those protocols is outside the scope of the specification.

## Use of javax.rmi.PortableRemoteObject.narrow

Use of the `javax.rmi.PortableRemoteObject.narrow` is a requirement and there to ensure CORBA can be used as the underlying protocol.

# EJB 1.0 and Industry Usage

The EJB 1.0 spec was released in March of 1998 and naturally by late 1999 we literally had implementations doing *all* of the above; JBoss used RMI (JMRP).  WebSphere used CORBA (IIOP); WebLogic used it's own custom T3 protocol; OpenEJB came along shortly after and also implemented a custom protocol.

As mentioned in EJB 1.0 all method invocations on an EJB are remote, so naturally there is a great deal of competition on invocation speed.  There are also new developers learning to write serailizable apps and getting things wrong.  Users start digging into the guts of some of these protocols and application portability becomes a bit of a problem.

# EJB 1.1

## RMI and CORBA requirements merged

As EJB 1.0 had effectively three separately handled remoting scenarios (RMI, CORBA, Other) and portability issues occured, a large focus of EJB 1.1 was to unify rules into one canonical set that covered all scenarios.  The term "RMI-IIOP" is introduced to serve that purpose. Sentences that previously read "The method argument and return value types must be legal types for Java RMI" now read "The method argument and return value types must be legal types for RMI-IIOP."  Sometimes the phrase "RMI/IIOP" is used as in "The method arguments must be legal types for RMI/IIOP"

There are now only 8 uses of the phrase "Java RMI" and only 7 references to plain "IIOP".  All other references now say "RMI-IIOP"

## Still Protocol Agnostic

The section titled "Support for other protocols" is deleted and not present, however IIOP is still not a requirement and vendors still have a choice.  Now there is a very brief chapter titled "Support for Distribution" that contains the following:

    The EJB specification further constrains the Java RMI types that
    can be used by enterprise beans to the legal RMI-IIOP types[7].
    This makes it possible for the EJB Container implementors to use
    RMI-IIOP as the object distribution protocol.  Note: The EJB 1.1
    specification does not require Container vendors to use
    RMI-IIOP. A later release of the J2EE platform is likely to
    require a J2EE platform implementor to implement the RMI-IIOP
    protocol for EJB interoperability in heterogeneous server
    environments.

## Duality of RMI-IIOP

The fact that we did a find replace for "Java RMI" to "RMI-IIOP" and "IIOP" to "RMI-IIOP" leaves the the term ambiguously refering to both requirements for being protocol agnostic and refering to the actual IIOP protocol.  Here are simple examples of each:

 - As application requirements: "An entity with bean-managed persistence can use any
RMI-IIOP Value Type as its primary key type; the primary key type of an entity with container-managed
persistence is more constrained."

 - As a protocol: "Note: Programs using the cast operator for narrowing the remote and home interfaces are likely to fail
if the Container implementation uses RMI-IIOP as the underlying communication transport."

This is because RMI is itself somewhat of a protocol abstraction -- JMRP is the actual protocol.  So a find-replace on "Java RMI" to "RMI-IIOP" creates a somewhat confusing result.  In the context of former "Java RMI" references the updated sentences now reading "RMI-IIOP" will detail hard requirements for being protocol agnostic.  In the context of sentences that use to refer to IIOP it will optional requirements that are protocol-specific.

### Trouble Eliminating RMI-IIOP

This makes eliminating anything that uses the phrase "RMI-IIOP" quite difficult.  Such a task requires a scalple, not an axe.

For example let's imagine "Java RMI" is world peace and "IIOP" is world hunger.  The 1.0 specification had a dozens of references and requirements around world peace and a handful of references to world hunger.  These then got combined to "PEACE-HUNGER" and left that way for several years.  We then decide later we wanted to abandon requirements for world hunger.  If we go about that by simply deleting all requirements that use the phrase "PEACE-HUNGER" we are not just eliminating world hunger we're also unintentionally deleting world peace.

## Use of javax.rmi.PortableRemoteObject.narrow

Use of the `javax.rmi.PortableRemoteObject.narrow` is still there as a requirement to ensure CORBA can be used as the underlying protocol.

# EJB 2.0

In EJB 2.0 is when the use of CORBA becomes an actual requirement.  The brief section from EJB 1.1 titled "Support for Distribution" is retitled "Support for Distribution and Interoperability" and goes from 2 pages to 24 pages.  However, only 12 new occurrences of the phrase "RMI-IIOP" are added bringing the spec from 38 usages of the term to 50.

## Introduction of the term "COBRA/IIOP"

To make matters more interesting the additional 12 uses of "RMI-IIOP" are not in the new "Support for Distribution and Interoperability" chapter.  This new chapter introduces the phrase "CORBA/IIOP" in an attempt to distingish itself from "RMI-IIOP" which is used heavily to refer to an abstraction.  For example:

    This section describes the interoperable mechanisms that support
    secure invocations on enterprise beans in intranets. These
    mechanisms are based on the CORBA/IIOP protocol.
    
The occurences in the specification that said "RMI-IIOP protocol" are left as-is and it is up to the reader to understand that "CORBA/IIOP" and "RMI-IIOP" are not two separate protocols.  This also makes "RMI-IIOP" a little more ambiguous.

## Still Protocol Agnostic

The same phrasing from EJB 1.1 is still present in the EJB 2.0 specification.

    The EJB specification further constrains the Java RMI types that
    can be used by enterprise beans to be legal RMI-IIOP types [7].
    This makes it possible for EJB Container implementors to use
    RMI-IIOP as the object distribution protocol.

However, the note afterwards stating servers do not need to implement RMI-IIOP is now gone.  Additionally, this note is in the section titled "Support for Distribution and Interoperability", which was deleted from the Jakarta Enterprise Beans 4.0 specification and likely should return.

This further proves how difficult it is to get rid of the presence of CORBA without accidentally going too far and removing things we still need.  In this case, the above phrasing is our only cary-over from EJB 1.0 that gave us any clue that the specification is attempting to be protocol agnostic.

# Conclusion

The above is an attempt at providing enough context to understand the evolution and nuances inherently in the overloaded term "RMI-IIOP" and why that leads to incredible difficulty in handling the removal of CORBA.

From the beginning EJB has said "you don't have to use CORBA, but it is possible and here is how."  It's been our general attempt in the Jakarta Enterprise Beans project to rollback requirements to the EJB 1.1 level which is the last version that did not require CORBA while still having sufficient language to ensure application developers did not do things to make CORBA  support  impossible.  

It's important that when we discuss removal of CORBA everyone invovled in the conversation can clearly see this line.  Anything that is CORBA-enabling must stay.  Anything that is CORBA-requiring must go.

There is a further complication that `javax.rmi.PortableRemoteObject` was removed from Java 11 and this API has been part of the CORBA-enabling set of requirements since EJB 1.0.  This was not factored into our Jakarta Enterprise Beans 4.0 release plan and likely needs some community discussion and prototyping to see if this can be handled in a Jakarta Enterprise Beans 4.1.

starksm64:
> It's important that when we discuss removal of CORBA everyone involved in the conversation can clearly see this line. Anything that is CORBA-enabling must stay. Anything that is CORBA-requiring must go.

If the point of leaving PortableRemoteObject in is simply an expediency step to minimize the amount of work needed to finalize EE 9, then that is one thing. This is still problematic as even the non-IIOP behavior of the JDK 8 PortableRemoteObject.narrow() call ends up relying on doing a simple cast inside of the com.sun.corba.se.impl.javax.rmi.PortableRemoteObject in OpenJDK based distros, and probably requires an implementation of the javax.rmi.CORBA.PortableRemoteObjectDelegate interface in others since javax.rmi.PortableRemoteObject relies on a delegate for any functionality. 

If the argument is that PortableRemoteObject is the right way to handle client interface to transport binding, then I disagree. The proper way to deal with this is via CDI providers and injection of the client view of the object. The EE 9 RMI-IIOP section should be stating that.

dblevins:
> > It's important that when we discuss removal of CORBA everyone involved in the conversation can clearly see this line. Anything that is CORBA-enabling must stay. Anything that is CORBA-requiring must go.
> 
> If the point of leaving PortableRemoteObject in is simply an expediency step to minimize the amount of work needed to finalize EE 9, then that is one thing. This is still problematic as even the non-IIOP behavior of the JDK 8 PortableRemoteObject.narrow() call ends up relying on doing a simple cast inside of the com.sun.corba.se.impl.javax.rmi.PortableRemoteObject in OpenJDK based distros, and probably requires an implementation of the javax.rmi.CORBA.PortableRemoteObjectDelegate interface in others since javax.rmi.PortableRemoteObject relies on a delegate for any functionality.
> 
> If the argument is that PortableRemoteObject is the right way to handle client interface to transport binding, then I disagree. The proper way to deal with this is via CDI providers and injection of the client view of the object. The EE 9 RMI-IIOP section should be stating that.

We should evaluate removing the requirement for `PortableRemoteObject.narrow` in a Jakarta Enterprise Beans 4.1.  In EJB 3.0 we added new-style remote interfaces that did not need to subclass from `java.rmi.Remote` or use `PortableRemoteObject.narrow`.  These are supported by everyone and are not marked optional, so theoretically we have reasonable evidence that all implementations can be made to function without `PortableRemoteObject.narrow` and there is hope it can be completely eliminated.

A challenge in testing/verifying this is most (all?) implementations of EJB have an optimization to skip the protocol layer if it is determined that something invoking the Remote interface of an EJB is in the same JVM.  In these situations EJB containers will "fake" a remote call's pass-by-copy semantics call by serializing then immediately deserializing the method arguments in memory before invoking the bean's method and doing a similar in-memory copy afterwards on the return value.  In these scenarios CORBA/IIOP or sockets of any kind are not involved and there is no need for `PortableRemoteObject.narrow`.  Unfortunately, this also means any TCK tests that might test a server's ability to handle a remote call without `PortableRemoteObject.narrow` may falsely pass if the test doesn't force a true remote scenario and the container therefore uses the optimized path.  This brings up another bit of work which is to take a close look at how we test remote interfaces to see if we are ensuring a truly remote scenario.

gerdogdu:
I am a little bit late on this but I really don't get why we have to maintain old RMI-IIOP tests. I really don't care which application server uses which protocol (and how) to implement EJB remoting protocol. We definitely remove this old technology and related tests from specification and related TCKs. We will also not support distributed interoperability.

>  In my opinion none of the RMI/IIOP APIs should remain since there are no useful required tests for them and really no point without a standard interop protocol. If a future version does introduce a standard remoting protocol, we can deal with the requirements at that point.

+++1 for this. We are trying to simplify the TCK for new incomers.
Regards.
Gurkan

kwsutter:
Pretty much agree with @gerdogdu's [comment](https://github.com/eclipse-ee4j/jakartaee-platform/pull/249#issuecomment-716801949).  The point of removing the Distributed Interop chapter from the EJB Spec was to lower the bar of entry and to simplify the testing.  Reading through the history was educational, but new users and implementors of Jakarta EE 9 won't care about the history.  If they want the history, they can refer to past versions of the Spec.  We just need to outline the requirements for CORBA and/or RMI/IIOP going forward.

If we need to keep the `PortableRemoteObject.narrow` for now just to allow progress and completion for Jakarta EE 9, I'm okay with that.  What I'm hoping we can accomplish at tomorrow's Platform call is to determine what we need to leave and what we can remove from the Platform Spec due to the removal of the Distributed Interop section of the EJB spec.

You also had this in your writeup:
`However, the note afterwards stating servers do not need to implement RMI-IIOP is now gone. Additionally, this note is in the section titled "Support for Distribution and Interoperability", which was deleted from the Jakarta Enterprise Beans 4.0 specification and likely should return.`

I hope you are not indicating that the whole Distributed Interop section needs to be returned...

Bottom line:  I'm now better educated on the history of these terms, but I don't feel any closer to resolving the issue of how does the removal of the Distributed Interop section of the EJB spec affect the Platform spec.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
 
 ===== RMI-JRMP
 
-*TODO Note:* What to do with this section? Leave as-is, mark Optional, remove?
-
```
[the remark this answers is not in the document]
kwsutter: After reading the history lesson, I'm now okay with the removal of this TODO.  All that this section does is briefly define what RMI-JRMP is.  We can leave it.

kwsutter:
> > We currently do not have significant tests for these guarantees. These tests are not there because they were part of the RMI-IIOP interoperability tests, which are removed as they are unfortunately also CORBA specific. We would need a new suite that is exclusively focused on portability, not interoperability. Should we find ourselves in a situation where new protocols find their way into the next generation of Jakarta EE implementations and people notice portability in serialization/deserialization semantics (a deserialized object graph is not quite the same when porting from one server to another), we will need to expand our tests.
> 
> @dblevins we also have other RMI-IIOP/Corba tests that we are waiting for input on removing. As mentioned on https://www.eclipse.org/lists/ejb-dev/msg00136.html, instead of making the RMI-IIOP + Corba tests optional, we are removing them to be consistent with how we handled other pruned/removed technologies. I'm bringing attention to this conversation in this thread so others can follow that conversation as it relates to whatever we decide here. Feedback on this other linked conversation is very much welcome in the ejb-dev mailing list thread (not here, unless it becomes important to this pull request). :-)

Scott, could you update this remark to indicate the current testing strategy around the use of the CORBA and RMI-IIOP property buckets?

scottmarlow:
> > > We currently do not have significant tests for these guarantees. These tests are not there because they were part of the RMI-IIOP interoperability tests, which are removed as they are unfortunately also CORBA specific. We would need a new suite that is exclusively focused on portability, not interoperability. Should we find ourselves in a situation where new protocols find their way into the next generation of Jakarta EE implementations and people notice portability in serialization/deserialization semantics (a deserialized object graph is not quite the same when porting from one server to another), we will need to expand our tests.
> > 
> > 
> > @dblevins we also have other RMI-IIOP/Corba tests that we are waiting for input on removing. As mentioned on https://www.eclipse.org/lists/ejb-dev/msg00136.html, instead of making the RMI-IIOP + Corba tests optional, we are removing them to be consistent with how we handled other pruned/removed technologies. I'm bringing attention to this conversation in this thread so others can follow that conversation as it relates to whatever we decide here. Feedback on this other linked conversation is very much welcome in the ejb-dev mailing list thread (not here, unless it becomes important to this pull request). :-)
> 
> Scott, could you update this remark to indicate the current testing strategy around the use of the CORBA and RMI-IIOP property buckets?

Update now via https://github.com/eclipse-ee4j/jakartaee-platform/pull/249#issuecomment-702873389

dblevins:
> Pretty much agree with @gerdogdu's [comment](https://github.com/eclipse-ee4j/jakartaee-platform/pull/249#issuecomment-716801949). The point of removing the Distributed Interop chapter from the EJB Spec was to lower the bar of entry and to simplify the testing. Reading through the history was educational, but new users and implementors of Jakarta EE 9 won't care about the history. If they want the history, they can refer to past versions of the Spec. We just need to outline the requirements for CORBA and/or RMI/IIOP going forward.

I think everyone agrees with @gerdogdu as that's been the goal from the start.  It's been difficult getting everyone to see that.

We aren't removing any that relates to CORBA in any capacity, we're simply removing the interop requirement.  Similarly, we need the spec to still say "you can implement any protocol you want" and this is the opening sentence in the PR.

The history is simply an attempt to break the circles we've gone in trying to resolve this topic.

> If we need to keep the `PortableRemoteObject.narrow` for now just to allow progress and completion for Jakarta EE 9, I'm okay with that. What I'm hoping we can accomplish at tomorrow's Platform call is to determine what we need to leave and what we can remove from the Platform Spec due to the removal of the Distributed Interop section of the EJB spec.

Agree.
 
> You also had this in your writeup:
> `However, the note afterwards stating servers do not need to implement RMI-IIOP is now gone. Additionally, this note is in the section titled "Support for Distribution and Interoperability", which was deleted from the Jakarta Enterprise Beans 4.0 specification and likely should return.`
> 
> I hope you are not indicating that the whole Distributed Interop section needs to be returned...

Correct.  Simply a note that we lost the last few sentences in the Enterprise Beans specification that said "we don't care what protocol you use."   We need some form of those sentences back in the Enterprise Beans spec itself.  I'd really prefer to leave that for a 4.1 and not try to do it in the 4.0.

If this PR is merged we would be covered as the opening sentence defines RMI-IIOP as protocol independent.  That text has been there for 20 years and is a key reason we can't simply delete this chapter from the Platform.

We could move it into the Enterprise Beans specification, but I wouldn't feel comfortable doing that now.

> Bottom line: I'm now better educated on the history of these terms, but I don't feel any closer to resolving the issue of how does the removal of the Distributed Interop section of the EJB spec affect the Platform spec.

We need people comfortable enough to merge this PR or suggest better language.

If anyone wants to suggest better language I recommend reading the two-paragraph "RMI-IIOP" section on page 19 of the J2EE 1.2 specification.  This is the original text that grew into the RMI-IIOP chapter in our current Jakarta EE 8 spec.

This PR attempts to roll this RMI-IIOP chapter back to effectively "J2EE 1.2 specification" level while also being clear there are requirements that used to be here and are deleted.

If I were to make any change based on the last few weeks of conversation, it'd be to add a note that `PortableRemoteObject.narrow` is going away.  If we plan to move this section to the Enterprise Beans specification in a future revision, I'd probably want to note that too so people know where to look going forward.

gerdogdu:
> This PR attempts to roll this RMI-IIOP chapter back to effectively "J2EE 1.2 specification" level while also being clear there are requirements that used to be here and are deleted.

I don't think that it is necessary. We are starting a new chapter in Jakarta EE and in my opninion no need to do like this. Just remove the complete RMI-IIOP and related tests from TCK is enough. Why do you need to emphasize on J2EE 1.2 specification? Who cares...

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
```diff
@@ -1000,91 +1000,48 @@ _https://jakarta.ee/specifications/xml-web-services/_ .
...
+narrow method of javax.rmi.PortableRemoteObject) when accessing
+remote Enterprise JavaBeans components, as described in the Jakarta Enterprise
+Beans 4.0 specification.  This allows Enterprise Beans and their clients to be
+protocol independent and portable to Jakarta EE implementations that may use
+CORBA or RMI as their underlying protocol.  However, there is currently no
+requirement that Enterprise Beans be accessible using the IIOP protocol itself
+or that Jakarta EE implementations support any form of distributed interoperability
+with other Jakarta EE implementations.
+
+The use of the term RMI-IIOP has existed since J2EE 1.2 and EJB 1.1, predates
```
kwsutter, line 1031: This paragraph reads like it should be a "note" and properly set away from the main flow of the text.  I'm okay with highlighting this educational history, but let's set it apart from the key message.

kwsutter:
@dblevins After re-reading a clean version of your proposed update to this single section, I'm pretty much okay with it.  I just added one comment about setting the history paragraph separate from the main text flow.  But, I think it's okay for Jakarta EE 9.  Of course, we still have several other sections with potential implications that still need to be resolved...

kwsutter:
> In my opinion none of the RMI/IIOP APIs should remain since there are no useful required tests for them and really no point without a standard interop protocol. If a future version does introduce a standard remoting protocol, we can deal with the requirements at that point.

This approach would definitely simplify things for Jakarta EE 9.  I just don't know if this approach would satisfy the requirements for remote EJBs.  @dblevins?

> > This PR attempts to roll this RMI-IIOP chapter back to effectively "J2EE 1.2 specification" level while also being clear there are requirements that used to be here and are deleted.
> 
> I don't think that it is necessary. We are starting a new chapter in Jakarta EE and in my opninion no need to do like this. Just remove the complete RMI-IIOP and related tests from TCK is enough. Why do you need to emphasize on J2EE 1.2 specification? Who cares J2EE 1.2 any more?

I agree that we shouldn't emphasize the J2EE 1.2 specification any longer.  But, after reading through a clean version of David's proposed update (without all of these intervening comments), I think the reference to J2EE 1.2 is okay and sufficient for the Jakarta EE 9 update.

Eventually, this area needs to be addressed again when we decide to remove the dependency on `PortableRemoteObject` and then we can go at it with a butcher knife.  We're trying to be pragmatic with what we can or should do with Jakarta EE 9 and what needs to be put off to another day.

dblevins:
> > In my opinion none of the RMI/IIOP APIs should remain since there are no useful required tests for them and really no point without a standard interop protocol. If a future version does introduce a standard remoting protocol, we can deal with the requirements at that point.
> 
> This approach would definitely simplify things for Jakarta EE 9. I just don't know if this approach would satisfy the requirements for remote EJBs. @dblevins?

Use of javax.rmi is planned for removal in Enterprise Beans 4.1, however we still heavily use the java.rmi.* classes such as java.rmi.Remote, java.rmi.RemoteException, etc.  These are still in Java 11 and there is currently no plan to remove them from the JVM or their use in the Enterprise Beans specification as they are not specific to CORBA/IIOP and do enable use of any protocol.

dblevins:
@kwsutter remaining TODOs should be gone
@starksm64 I've listed some other protocol examples to be more clear on spirit

kwsutter:
@dblevins I'm re-reviewing this whole section.  In the subsection titled "Removed Jakarta Technologies", there is a row to remove "CORBA, Java IDL, RMI-IIOP" in Jakarta EE 9.  Should that row be removed.  (It was most likely added by yours truly...)  I just want to make sure that we're on the same page.  Thanks.

dblevins:
@kwsutter I've gone ahead and yanked that row from the table.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
kwsutter: @dblevins This should be corrected.

specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
kwsutter: @dblevins This should be corrected.

kwsutter (approved):
@dblevins There are a couple of minor items that other people have pointed out in their reviews that were not updated yet.  I added replies indicating my agreement with those easy changes.  Other than that, I'm okay with going with what you have provided.  As stated on today's call, we have more work ahead of us in this area...  But, we can only bite off so much for Jakarta EE 9.

kwsutter:
This PR now seems consistent with what was discussed on the Platform call today (10/27).  I will go ahead with the merge so that we have a fresh version to work with.  We still have time to adjust the wording if someone has additional suggestions or updates.

PR #255, opened 2020-10-27
Update Interop chapter of Platform Spec

kwsutter:

This is a short chapter, but it was filled with "requirements" of supporting all of these various protocols.  Some are related to the CORBA and IIOP things we've been discussing, but it also talks about generic interop via data formats and networking protocols.  So, we can't just cut the whole chapter.  I've tried to soften the requirements on CORBA and IIOP.  Looking for comments if I have went far enough...  Thanks!

### DIFF
--- a/specification/src/main/asciidoc/platform/Interoperability.adoc
+++ b/specification/src/main/asciidoc/platform/Interoperability.adoc
@@ -2,7 +2,7 @@
 == Interoperability
 
 This chapter describes the interoperability
-requirements for the Jakarta™ EE Platform.
+features for the Jakarta™ EE Platform.
 
 === Introduction to Interoperability
 
@@ -23,7 +23,7 @@ platform, or Unix® workstation.
 * standalone Java™ technology-based applications
 that are not directly supported by the Jakarta EE platform.
 
-It is the interoperability requirements of the
+It is the interoperability features of the
 Jakarta EE platform, set out in this chapter, that make it possible for it
 to provide indirect support for various types of clients, different
 hardware platforms, and a multitude of software applications. The
@@ -31,33 +31,31 @@ interoperability features of the Jakarta EE platform permit the underlying
 disparate systems to work together seamlessly, while hiding much of the
 complexity required to join these pieces together.
 
-The interoperability requirements for the
-current Jakarta EE platform release allow:
+The interoperability features for the
+current Jakarta EE platform release may allow:
 
 * Jakarta EE applications to connect to legacy
-systems using CORBAfootnote:removed9[Removed from Jakarta EE 9.] or low-level socket interfaces.
-* Jakarta EE applications to connect to other Java
-Jakarta applications across multiple Jakarta EE products, whether from different
-Product Providers or from the same Provider, and multiple Jakarta EE
-platforms.
+systems using CORBA or low-level socket interfaces.
+* Jakarta EE applications to connect to other
+Jakarta EE applications across multiple Jakarta EE products and platforms.
 
 In this version of the specification,
 interoperability between Jakarta EE applications running in different
-platforms is accomplished through the HTTP protocol, possibly using SSL,
-or the Jakarta Enterprise Beans interoperability protocol based on IIOPfootnote:removed9[].
+platforms may be accomplished through the HTTP protocol, possibly using SSL,
+or the Jakarta Enterprise Beans interoperability protocol based on IIOP.
 
 === Interoperability Protocols
 
-This specification requires that a Jakarta EE
-product support a standard set of protocols and formats to ensure
+This specification allows a Jakarta EE
+product to support a standard set of protocols and formats to ensure
 interoperability between Jakarta EE applications and with other
 applications that also implement these protocols and formats. The
-specification requires support for the following groups of protocols and
+specification supports the following groups of protocols and
 formats:
 
 * Internet and web protocols
-* OMG protocolsfootnote:removed9[]
-* Java technology protocols **TODO** JRMP may also be removed?
+* OMG protocols (optional)
+* Java technology protocols
 * Data formats
 
 Most of these protocols and formats are
@@ -90,7 +88,7 @@ are required as part of the Jakarta Enterprise Beans interoperability protocol i
 specification.
 * SOAP 1.1—SOAP is a presentation layer
 protocol for the exchange of XML messages. Support for SOAP layered on
-HTTP is optional, as described in the Jakarta XML-based RPCfootnote:removed9[] and 
+HTTP is optional, as described in the Jakarta XML-based RPCfootnote:removed9[Removed in Jakarta EE 9] and 
 Jakarta XML Web Servicesfootnote:optional9[Made optional in Jakarta EE 9] specifications.
 * SOAP 1.2—SOAP 1.2 is the version of the SOAP
 protocol standardized through W3C and supported by Jakarta XML Web Servicesfootnote:optional9[].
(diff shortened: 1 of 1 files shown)


dblevins:
@kwsutter thanks for the edits, but... why are we keeping all this old stuff?
(sorry, couldn't resist.  everyone else got a turn :) )

specification/src/main/asciidoc/platform/Interoperability.adoc
```diff
@@ -2,7 +2,7 @@
 == Interoperability
 
 This chapter describes the interoperability
-requirements for the Jakarta™ EE Platform.
+features for the Jakarta™ EE Platform.
```
dblevins, line 5: At first this appeared to me as noise, but after a few moments I think it's clear you're attempting to eliminate the blanket statement "everything in this chapter is a requirement", so you could dial things in more specifically within the chapter.  Did I get that right?

dblevins (approved):
This looks as accurate as we can get for our current state

kwsutter:
> @kwsutter thanks for the edits, but... why are we keeping all this old stuff?
> (sorry, couldn't resist. everyone else got a turn :) )

Fair question.  This chapter talks about more protocols than the ones we want to get rid of eventually.  Now, we the other valid protocols are sufficiently covered elsewhere, then maybe this chapter could be removed.

specification/src/main/asciidoc/platform/Interoperability.adoc
kwsutter: Yes, that's what I was trying to do...

kwsutter:
Thanks for the quick review and approval, @dblevins.  I'll leave it up open for a bit to see if I get any other comments.

specification/src/main/asciidoc/platform/Interoperability.adoc
scottmarlow: Are there any parts of (now pruned/removed) `interoperability` that we need to specify for Jakarta EE 9?  

Compatible implementations may support RMI-IIOP + Corba but might not implement CSIv2, so application developers may get confused by seeing a chapter on `interoperability` in the EE 9 Platform spec.

I will have more feedback than I added below if we do keep this chapter.

specification/src/main/asciidoc/platform/Interoperability.adoc
```diff
@@ -23,41 +23,39 @@ platform, or Unix® workstation.
...
+systems using CORBA or low-level socket interfaces.
+* Jakarta EE applications to connect to other
+Jakarta EE applications across multiple Jakarta EE products and platforms.
 
 In this version of the specification,
 interoperability between Jakarta EE applications running in different
-platforms is accomplished through the HTTP protocol, possibly using SSL,
-or the Jakarta Enterprise Beans interoperability protocol based on IIOPfootnote:removed9[].
+platforms may be accomplished through the HTTP protocol, possibly using SSL,
+or the Jakarta Enterprise Beans interoperability protocol based on IIOP.
```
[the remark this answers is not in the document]
kwsutter: I agree that this chapter is a little wonky.  But, until we can decide that we can remove the whole CORBA and RMI/IIOP requirement, I think we have to leave these soft requirements.  The only change that I am proposing is to soften the "must haves" to "could haves" or "should haves"...

specification/src/main/asciidoc/platform/Interoperability.adoc
kwsutter: There is another section in the Platform Spec that specifically addresses EJB 4.0 requirements.  The removal of the Distributed Interop was mentioned in that section.

This Interop chapter is wonky.  It brings in aspects of several types of interop -- network protocols, data formats, and the corba stuff.  We're trying to soften the Corba/OMG stuff and leave the other aspects of interop as-is.  Again, not perfect, but not sure what else to do at this point.  We're kind of in a limbo state...

kwsutter:
One easy change would be to change the title of this chapter to "Application Interoperability" -- just to help differentiate between the information in this chapter and the "distributed interop" topic.

specification/src/main/asciidoc/platform/Interoperability.adoc
scottmarlow: @kwsutter If we could avoid using the single `Interoperability` word as much, as it is overloaded:

1. Replace title with `Application Interoperability` or `App Interop`.
1. Replace `the Jakarta Enterprise Beans interoperability protocol based on IIOP` with `Corba`.
1. Replace `the Jakarta Enterprise Beans interoperability protocol in the Jakarta Enterprise Beans specification` with `Corba`.

We still would use `interoperability` in the generic sense for elsewhere in this document.

scottmarlow:
> One easy change would be to change the title of this chapter to "Application Interoperability" -- just to help differentiate between the information in this chapter and the "distributed interop" topic.

https://github.com/eclipse-ee4j/jakartaee-platform/pull/255#discussion_r513594821 has that and one additional suggestion. :-)

kwsutter:
Since Ch 10 (Distributed Interop) was removed from the EJB Spec, the corresponding references from this section of the Platform spec needed to be removed as well.  I think this reads better now.  I will ask for a re-review from @scottmarlow and @dblevins.  But, will accept comments from anyone.  Thanks!

scottmarlow (approved):

specification/src/main/asciidoc/platform/Interoperability.adoc
```diff
@@ -1,8 +1,8 @@
 [[a2845]]
-== Interoperability
+== Application Interoperability
 
-This chapter describes the interoperability
-requirements for the Jakarta™ EE Platform.
+This chapter describes the application interoperability
+features for the Jakarta™ EE Platform.
 
 === Introduction to Interoperability
```
Emily-Jiang, line 7: Should this title to be updated to `Introduction to Application Interoperability` to sync up with the previous title change?

specification/src/main/asciidoc/platform/Interoperability.adoc
dblevins: The addition of the "application interoperability" I think has good intent, but not the results we're hoping for and our best bet is to leave it as-is.

 - We haven't said how "application interoperability" is the same as or different than "interoperability" which is still heavily used in this section and elsewhere.  As someone who just went on a 22-year dig through an IIOP, RMI, RMI/IIOP, RMI-IIOP, CORBA/IIOP, CORBA-IIOP, CORBA, RMI-JMRP deciphering journey I have heightened empathy for trying to reverse engineer intent behind all the variations.  Sometimes they are synonyms, sometimes antonyms.
 - There were phases in EJB where terms were renamed and not carried out consistently, leaving the old term hanging around, no explanation if it was the same or different and therefore making the topic less clear, not more clear.  @Emily-Jiang points this out on one other title, but even changing that one title does not eliminate the heavy reliance on the bare word "interoperability."  Unless we change them all, we are just adding more "interoperability" confusion, not reducing confusion.
 - The word "interoperability" is used in the chapter near the nouns client, server, protocol, legacy system, and application.  Many of us would consider selecting "application" as the primary description to be odd.  In other areas we deliberately attempt to describe client, server and application as three distinct things,

We avoid the least confusion by just leaving it as is.  If we do want to rename it, we'd need to all agree it's the right new term, do it consistently throughout the spec so the old term is left hanging undefined or undifferentiated, and explain the motivation for the change in the changelog.

I don't think any of this is worth it.

dblevins: So there's something encouraging, huge +1 to remove "Jakarta Enterprise Beans interoperability protocol based on IIOP" for the reasons above.  We definitely do *not* need another undefined term to describe remote EJB capabilities.

kwsutter: Since we cleared up some of the IIOP items, then changing this chapter back to just "Interoperability" is fine with me.  I'll revert that change.

specification/src/main/asciidoc/platform/Interoperability.adoc
kwsutter: As David pointed out, that proposed title change does have a ripple effect as you discovered.  I'm going to change it back.

kwsutter:
> > One easy change would be to change the title of this chapter to "Application Interoperability" -- just to help differentiate between the information in this chapter and the "distributed interop" topic.
> 
> [#255 (comment)](https://github.com/eclipse-ee4j/jakartaee-platform/pull/255#discussion_r513594821) has that and one additional suggestion. :-)

Based on @dblevins' input, we have decided to revert back to the original title of "Interoperability".  Too much of a ripple effect across the spec, and we haven't differentiated between "app interop" and "interop".

kwsutter:
I think we have sliced and diced this section sufficiently.  I will go ahead with the merge.  Additional changes can be proposed via separate PRs.  Thanks!

PR #260, opened 2020-10-29
Backwards compatibility updates for Jakarta EE 9

kwsutter:

We had an existing "Compatibility and Migration" chapter, so I'm just adding a subsection about "Backwards Compatibility".  It's not quite done yet.  I'll take comments on the initial commit, but there is more to come.

### DIFF
--- a/specification/src/main/asciidoc/platform/CompatibilityMigration.adoc
+++ b/specification/src/main/asciidoc/platform/CompatibilityMigration.adoc
@@ -30,18 +30,49 @@ applications written to a previous version of the platform will continue
 to work without change and with identical behavior on the current
 version of the platform.
 
+[[a3901]]
+==== Backwards Compatibility for Jakarta EE 9
+
+The Backwards Compatibility Requirements for Jakarta EE Specifications is documented here: 
+_https://eclipse-ee4j.github.io/jakartaee-platform/CompatibilityRequirements_.
+Unfortunately, some of these requirements were not able to be adhered to for Jakarta EE 9.
+This was due to the required namespace change from `javax` to `jakarta`.
+
+===== `javax` to `jakarta` Namespace Change
+Per the Jakarta EE 9 Release Plan (https://eclipse-ee4j.github.io/jakartaee-platform/jakartaee9/JakartaEE9ReleasePlan), the whole `javax` namespace needed to change to `jakarta`.
+This had a ripple effect across all of the Specification documents, the APIs, the Javadocs, the TCKs (Technology Compatibility Kits), and the CIs (Compatible Implementations).
+In turn, this required change will also affect applications.
+Thus, the requirement that changes _"must be source-code compatible, binary compatible, and behavior compatible for applications"_ can not be honored for Jakarta EE 9.
+
+Compatible Implementations, IDEs, and other tools are providing mechanisms to make this change as painless as possible.
+One such effort is the Eclipse Transformer Project (https://projects.eclipse.org/projects/technology.transformer).
+This framework aids with the binary transformation of Java(TM) class files -- replacing the `javax.*` package references to the corresponding `jakarta.*` packages.
+In addition to being used with several of the Jakarta EE 9 Compatible Implementations, the Transformer has also been used to transform test buckets and other utility libraries.
+
+====== Schemas
+blah, blah, blah
+
+====== Properties
+blah, blah, blah
+
+===== Removed Technologies
+blah, blah, blah
+
 === Migration
 
 Migration is the act of converting an
 application to use new facilities introduced in this release of the
 platform.
 Given the required move from the `javax` namespace to the `jakarta` namespace in
-Jakarta EE 9, some migration of an application may be required.
+Jakarta EE 9, some migration of applications may be required.
 Some application servers may provide additional facilities to make this migration 
 more consumable.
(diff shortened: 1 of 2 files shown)


ivargrimstad (approved):

scottmarlow (approved):

dblevins:
@kwsutter This is a great first start and has the right key content, but I think it needs an overhaul to be more in "spec tone" and be more terse.  I'll lock myself into a quiet place early tomorrow, come up with another potential draft text and then connect with you over slack so we can see if it fits with what you have in mind.  I'll set myself an aggressive deadline to have something by 11am pacific so there's enough time in the day on your end for us to hammer it all out.

dblevins:
Feel free to use as much or as little of this as you want.  Consider it an attempt to be helpful and no expectation all of this must be incorporated.  I think we need some sort of "Specification Author's Guidelines" and I saw this as an opportunity to take a step in that direction.

In writing this I wonder if having an editor would be a good use of our WG budget at least until we all find our feet.  We're very likely to have a spec document filled with many different writing styles.

---

Some thoughts on strengthening the opening:

    The Backwards Compatibility Requirements for Jakarta EE Specifications is
    documented here:
    _https://eclipse-ee4j.github.io/jakartaee-platform/CompatibilityRequirements_.
    Unfortunately, some of these requirements were not able to be adhered to for
    Jakarta EE 9.  This was due to the required namespace change from `javax` to
    `jakarta`.

Opening with a "go read this link" feels odd as specifications should ideally be more standalone.  It's also odd there's a two-paragraph definition of compatibility proceeding this and it's unclear how they relate.

We say "Unfortunately" making the tone sound like an apology.  Specification text should likely be free from words that add emotional tone; fortunate, unfortunate, fantastic, terrible, etc.

We say "some of these requirements were not able to be adhered to", but we aren't specific as to which ones.  Specifications should ideally be specific.

There's a statement later "must be source-code compatible, binary compatible, and behavior compatible for applications" which is more specific.  We don't say which of these we honored and which we could not, but this feels like the most valuable information to tell the user upfront.

Attempted rewrite:

    Due to the migration from the `javax` namespace to the `jakarta` namespace, Jakarta
    EE 9 is not source-code compatible or binary compatible with previous releases.
    However, Jakarta EE 9 is behavior compatible and all `jakarta` APIs present
    share identical method signatures and behavior with their `javax` equivalents.
    
    Applications migrating to Jakarta EE 9 from previous releases will require
    modification in the areas detailed in the sections below.

    
From here the flow of "Namespace", "Properties", "Schema" is a great order.

I would ensure these titles are consistent in if we do or do not use "change": pick either "Namespace Change", "Properties Changes", "Schema Changes" or plain "Namespace", "Properties", "Schema"

I would ensure these sections are consistent in giving the "from this to that" information.  The Schema section does not mention the previous namespace.  The Namespace and Properties do mention the previous namespace or property prefix.

I would drop any commentary from these sections such as "another first" or background information such "this had a ripple effect" and how we got here.  I think I need to revise my RMI-IIOP update to remove similar commentary.

I would ensure each section has "here's what you need to do" statement such as, "Application deployment descriptors using the prior 'http....' namespace must be updated to the new 'http....' namespace."

I'd move the migration tools description from the "Namespace" section into a dedicated "Migration Tools" section at the very end of "Backwards Compatibility for Jakarta EE 9."  I would state such tools are outside the scope of the specification and then give the information. 

I would not reference the Jakarta EE 9 Release Plan.  In my mind the release plans are there to describe the expected end result, they should not be part of the end result.  Any critical information from the release plans should be moved into the final spec, not referenced by the final spec.

I would replace "implementation-dependent" with either "implementation-specific" or "vendor-specific."  We use these two terms in a handful of places.  We likely should consistently use just one of those two and eliminate the other, but that's a change for another day.

### Writing style (just FYI)

You have a very distinct style of starting sentences with a phrase and comma like "In the past," "But now," "In most cases," "While in other cases,"  Ideally writing style doesn't change from author to author and the spec reads as if written by one person.  Rather than use phrases like "Because of the rain, the party was canceled" I'd go for straight "The party was canceled because of the rain."

Sometimes this is actually creating sentence fragments.  An example is "While in other cases, only a portion of a Specification is deemed removable" because the inversion, "Only a portion of a Specification is deemed removable while in other cases" isn't a complete sentence.

Another example of writing style that may not be very spec-like is "Another _first_ for Jakarta EE 9 is the removal of older, rarely-used technologies. This is consistent with the need to lower the bar of entry for new compatible implementations."  This is great for an article, but for a spec it should likely be more direct, boring and flat.

"The following technologies have been removed Jakarta EE 9 to lower the bar of entry for new compatible implementations."

kwsutter:
FYI, the Backwards Compatibility Requirements for Jakarta EE Specifications is documented here: 
_https://eclipse-ee4j.github.io/jakartaee-platform/CompatibilityRequirements_.

kwsutter:
Thanks, @dblevins, for the comments and suggestions.  Firming up the wording for a spec is an "art" that I have not spent much time with.  I appreciate the input.  I have updated the various sections based on your input.  If you have the time to re-review before Sunday, that would be great.  If not, I will probably go ahead with the merge so that I can an updated version of the Spec documents generated for a wider community review by Monday.  Thanks.

kazumura (approved):

PR #268, opened 2020-11-04
Eliminate commentary and history in RMI-IIOP description

dblevins:
Attempting to incorporate the same feedback I gave Kevin in https://github.com/eclipse-ee4j/jakartaee-platform/pull/260 

 - Focus more clearly on requirements
 - Removed any commentary about the requirements, including explanations of history

### DIFF
--- a/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
+++ b/specification/src/main/asciidoc/platform/ApplicationProgrammingInterface.adoc
@@ -1018,9 +1018,8 @@ application component types can be clients of RMI objects.
 The RMI-IIOP subsystem is composed of APIs that allow for the
 use of RMI-style programming that is independent of the underlying
 protocol.  Implementations of these APIs may support the Java SE native RMI
-protocol (JRMP), the CORBA IIOP protocol or any custom protocol that is
-compatible with the RMI programming restrictions (see the RMI-IIOP spec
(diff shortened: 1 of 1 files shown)


kwsutter (changes requested):
@dblevins Thanks!  I do like this tighter language.  I did find a couple of things that should be cleaned up.  In the opening paragraph of this RMI-IIOP section (~line 1022), there is mention of the RMI-IIOP specification.  But, I can't find any reference in the Platform spec that outlines where to find this spec.

In the original Jakarta EE 8 spec, there was this mention (that has since been deleted with the Jakarta EE 9 updates).

_"RMI-IIOP is defined by chapters 5, 6, 13, 15, and section 10.6.2 of the CORBA 2.3.1 specification, available at http://www.omg.org/cgi-bin/ doc?formal/99-10-07, and by the JavaTM Language To IDL Mapping Specification, available at http://www.omg.org/cgi-bin/doc?ptc/2000-01-06."_

Or, better yet, maybe we should just update the RelatedDocs section with one or both of these references?  We need to provide one or the other to complete this section.

And, a nit...  I'm a fan of the "oxford comma" and I'd prefer a comma after protocol on ~line 1021:  "..IIOP protocol, or any custom.."

kwsutter (approved):
LGTM.  Thanks, @dblevins!

PR #282, opened 2020-11-24
Add minutes for Nov 10 and Nov 17 platform call

ivargrimstad:
Signed-off-by: Ivar Grimstad <ivar.grimstad@eclipse-foundation.org>


keilw:
Hi,

Is there a call today or are there not enough People because of the Holidays (at least in America)?

Werner

PR #291, opened 2021-01-15
Split Jakarta JSON Processing, JSON Binding and WebSocket

Thihup:
I don't know if this is the correct way, but reading it can cause some confusion.

Should I add to the end of the list or can it be next to the Jakarta JSON Processing?

Signed-off-by: Thiago Henrique Hupner <thihup@gmail.com>


dmatej (approved):
Seems better to me :-)

ivargrimstad (approved):

kwsutter (approved):
Much clearer.  Thanks!

PR #297, opened 2021-01-26
[Issue 296] Drop SE 8 requirement as a justification for EE 9 not hav…

### LINKED ISSUE #296: Unneeded SE 8 support mention in JPMS module name discussion
Section 6.1.1.1 includes the following:

"Jakarta EE 9 is not defining a module naming convention since Jakarta EE 9 is only supporting Java SE 8."

EE 9.1 is about Java SE 11, so make that:

"Jakarta EE 9 does not define a module naming convention."

This issue could be part of #287, but that one is more complex in nature, whereas this seems like a simple text cleanup for which I'll send up a PR. Even in the context of EE 9.0 it's not necessary to explain why EE doesn't define a module naming convention.

bstansberry:
…ing a module naming convention

Signed-off-by: Brian Stansberry <brian.stansberry@redhat.com>

Resolves #296


ivargrimstad (approved):

kwsutter (approved):
Thanks, @bstansberry!

PR #340, opened 2021-04-27
Managed Beans version should be 2.1-SNAPSHOT

kwsutter:

Managed Beans is not being updated for Jakarta EE 9.1, so the version should be set to 2.1-SNAPSHOT for normal development builds.


kazumura (approved):
LGTM. Thanks.

PR #385, opened 2021-07-22
Prep GH Pages for new Jakarta EE 10 release

kwsutter:

Initialize our GH Pages for the Jakarta EE 10 release.

One question is whether we still need or want the Specifications.md page.  It's difficult to maintain (easy to get out of date).  It's also referencing the EE4J implementations, which is a little biased when we have Compatible Implementations for some of the specs outside of the EE4J domain.  Comments or suggestions on this page would be appreciated.


jakartaee9/JakartaEE9.1.md
```diff
@@ -24,10 +24,9 @@ No other individual component (feature) release plans are expected.
```
kazumura, line 24: remove "(work in progress)"

jakartaee9/JakartaEE9.1.md
```diff
@@ -24,10 +24,9 @@ No other individual component (feature) release plans are expected.
 
 Overall schedule dates are ***tentative*** and ***subject to change***.
```
kazumura, line 25: remove this line

kazumura (commented):
Thanks, @kwsutter .
I left some minor suggestions for "JakartaEE9.1.md".

kazumura:
Regading Specification.md page, I agree it is difficult to maintain.
And we can get almost same information from "https://jakarta.ee/specifications/".
So I would suggest to remove Specification.md.

kwsutter:
> Regading Specification.md page, I agree it is difficult to maintain.
> And we can get almost same information from "https://jakarta.ee/specifications/".
> So I would suggest to remove Specification.md.

Thanks for the review and "vote" for removal of Specifications.md page, @kazumura!  Barring any further comment to keep it, I will probably remove it.

jakartaee9/JakartaEE9.1.md
kwsutter: Will do.  Thanks!

jakartaee9/JakartaEE9.1.md
kwsutter: Good idea.  Thanks!

ivargrimstad (approved):

kwsutter:
> > Regading Specification.md page, I agree it is difficult to maintain.
> > And we can get almost same information from "https://jakarta.ee/specifications/".
> > So I would suggest to remove Specification.md.
> 
> Thanks for the review and "vote" for removal of Specifications.md page, @kazumura! Barring any further comment to keep it, I will probably remove it.

Decided to go ahead with the removal.  We already have a reference to the main https://jakarta.ee/specifications/ page and that's a better reference anyway.

kazumura (approved):

edbratt:
Rather than remove it, what about just editing the Specifications page so it requires less maintenance. The links to the GH repositories are harder to find from the Specifications normative pages. We could certainly update the Specification column target to the normative intermediate landing page (i.e. specifications/annotations (as opposed to specifications/annotations/2.0 as an example). I could take a stab at this and if the consensus still is, discard it, we can still do that.

kwsutter:
> Rather than remove it, what about just editing the Specifications page so it requires less maintenance. The links to the GH repositories are harder to find from the Specifications normative pages. We could certainly update the Specification column target to the normative intermediate landing page (i.e. specifications/annotations (as opposed to specifications/annotations/2.0 as an example). I could take a stab at this and if the consensus still is, discard it, we can still do that.

Something along the lines of what I just committed?  Seems like a good compromise.

edbratt:
@kwsutter Yeah, that looks good to me. Hopefully it won't require frequent updates. Thanks!

ivargrimstad:
> @kazumura and @ivargrimstad, since you both had already approved this PR, what do you think about this slightly modified Specifications.md page instead of just deleting it?

I like it!

kazumura:
> @kazumura and @ivargrimstad, since you both had already approved this PR, what do you think about this slightly modified Specifications.md page instead of just deleting it?

Looks better!

kwsutter:
Thanks for the quick reviews!  Per my comments on yesterday's Platform call, I also updated the expected release date for 1Q2022 in the schedule.

Issue #413, opened 2021-09-24
Refactor specification requirements such that JNDI is not required by Core Profile


starksm64:
* Is your feature request related to a problem? Please describe.
The JNDI namespace is not something that microservice and small application environments have a need of. Further, none of the Core Profile specifications make use of or define integration with JNDI. Integration with JNDI should not be a requirement for Core Profile implementations.

* Describe the solution you'd like
The platform specification needs to factor out requirements and exclude the Core Profile where applicable. The JNDI of the 
ApplicationProgrammingInterface.adoc needs to be updated to exclude the Core Profile from JNDI requirements.

* Describe alternatives you've considered
None.

* Additional context
None.

PR #433, opened 2021-11-18
First pass at removing core profile requirements, #413

starksm64:
Signed-off-by: starksm64 <starksm64@gmail.com>

### DIFF
--- a/specification/src/main/asciidoc/platform/Profiles.adoc
+++ b/specification/src/main/asciidoc/platform/Profiles.adoc
@@ -138,7 +138,7 @@ requirements.
 [[a3240]]
 === Requirements for All Jakarta EE Profiles
 
-The Java Platform, Standard Edition 8 is the
+The Java Platform, Standard Edition 11 is the
 required API foundation for any Jakarta EE profile.
 
 The following technologies are required to be
@@ -147,12 +147,6 @@ present in all Jakarta EE profiles:
 * Resource and component lifecycle
 annotations defined by the Common Annotations specification ( _Resource_,
  _Resources_, _PostConstruct_, _PreDestroy_ )
-
-The following functionality is required to be
-supported in all Jakarta EE profiles other than the Core Profile:
-
-* JNDI “java:” naming context (see <<a607, JNDI Naming Context>>)
-* Jakarta Transactions
 
 === Optional Features for Jakarta EE Profiles
 
--- a/specification/src/main/asciidoc/platform/Profiles.adoc
+++ b/specification/src/main/asciidoc/platform/Profiles.adoc
@@ -112,7 +112,7 @@ profile may not allow omitting a package or type or method from an API
 specified elsewhere, unless the specification for that API explicitly
 allows for this to happen.
 
-Although this specification does not define
+Although the Platform specification does not define
 any APIs, a profile may do so. Since such an API would be available only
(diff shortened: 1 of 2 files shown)


specification/src/main/asciidoc/platform/Profiles.adoc
```diff
@@ -9,6 +9,8 @@ The Jakarta EE Web Profile Specification,
 published in conjunction with the present specification, defines the
 first Jakarta EE profile, the Web Profile.
 
+The Jakarta EE Core Profile Specification is a minimal profile that is published separately from the Web Profile and this specification.
```
kwsutter, line 12: Just to be clear, should you say "..separately from the Web Profile and this **Platform** specification."

specification/src/main/asciidoc/platform/Profiles.adoc
```diff
@@ -147,7 +149,7 @@ annotations defined by the Common Annotations specification ( _Resource_,
  _Resources_, _PostConstruct_, _PreDestroy_ )
 
 The following functionality is required to be
-supported in all Jakarta EE profiles:
+supported in all Jakarta EE profiles other than the Core Profile:
```
kwsutter, line 152: I think this whole section "Requirements for All Jakarta EE Profiles" needs to be re-worked.  Just inserting the "other than Core Profile" phrase opens the door for other Profiles to also not follow what's listed in this section.  Maybe this section just only contain the Java SE requirements and Common Annotations?  And, the requirement for Transactions and java: naming context should be moved to the Web Profile?  Just an idea.  But, we need to do something more than just inserting the "other than Core Profile" phrase.

starksm64: The simplest would be to remove JNDI/JTA from this section and add it as required in the Web Profile.

specification/src/main/asciidoc/platform/Profiles.adoc
starksm64: Fixed

specification/src/main/asciidoc/platform/Profiles.adoc
```diff
@@ -136,7 +138,7 @@ requirements.
 [[a3240]]
 === Requirements for All Jakarta EE Profiles
 
-The Java Platform, Standard Edition 8 is the
+The Java Platform, Standard Edition 11 is the
 required API foundation for any Jakarta EE profile.
```
ederks85, line 142: I would replace the word "foundation" with "compile level". Foundation feels a bit too open for interpretation, where one could read it as compile and runtime level.

ederks85 (approved):

specification/pom.xml
```diff
@@ -205,44 +205,6 @@
                             </attributes>
                         </configuration>
                     </execution>
-                    <!-- ManagedBeans HTML generation -->
```
scottmarlow, line 208: I assume this is for moving managed beans to the core profile.

scottmarlow (approved):

specification/pom.xml
starksm64: The managed beans spec was moved to its own repo:
https://github.com/eclipse-ee4j/managed-beans-spec

It was deprecated for EE10 as I remember, so I'm not sure what we are doing with that spec.

ivargrimstad (approved):

Issue #457, opened 2022-01-11
Does there need to be a statement made about the jakarta.* namespace being inapproriate for applications?


starksm64:
Based on the platform dev thread "Jakarta TCK package naming convention":
https://www.eclipse.org/lists/jakartaee-platform-dev/msg03072.html

The only statement that exists regarding restrictions around package namespace is from the Servlet specification:

15.2.1. Web Application Class Loader
Servlet containers that are part of a Jakarta EE Platform product should not allow the application to
override Java SE or Jakarta EE platform classes, such as those in java.*, javax.*, and jakarta.*
namespaces, that either Java SE or Jakarta EE do not allow to be modified.

Specs also reference the Java Language Specification which has always called out the java.* and javax.* packages as special in the terms of license section:

"... (b) does not modify, subset, superset or otherwise extend the Licensor Name Space, or include
any public or protected packages, classes, Java interfaces, fields or methods within
the Licensor Name Space other than those required/authorized by the Specification
or Specifications being implemented; ..."

The question is, does there need to be a formal statement regarding the use of the jakarta.* package namespece, and associated requirements on TCK deployments?

Related:
https://github.com/eclipse-ee4j/jaxrs-api/issues/1081
https://www.eclipse.org/lists/eclipselink-dev/msg08182.html

scottmarlow:
Also related:
https://github.com/eclipse-ee4j/jsonp/pull/351

starksm64:
Additional feedback from the EF regarding usage of their trademarks including the jakarta package:

1. See the Eclipse Foundation Trademark Usage Policy which says
    * Proper Usage of Eclipse Foundation Trademarks
        - “4. Only Eclipse Projects are authorized to develop or maintain software packages that use Eclipse Foundation namespaces such as 'org.eclipse' in their namespace. An important use of an Eclipse Trademark is the 'org.eclipse' string used on all namespaces for Eclipse Projects. This naming convention is used to identify code that has been developed as part of an Eclipse Project.”
    * Proper Usage of Eclipse WG Trademarks
        - "Use of any namespaces created by a WG including but not limited to ‘jakarta’, 'org.locationtech' or 'org.polarsys' is not permitted by this Policy unless authorized in advance in writing by Eclipse."
2. Also see the EF TCK License which says 
    * "4.c. comply with any requirements stated in the Specification with regard to subsetting, supersetting, modifying or extending the Specification in any Product claimed to be compatible with the Specification." 
[...]

PR #485, opened 2022-05-05
Add a statement about restrictions of using the jakarta package names…

starksm64:
…pace, #457

Signed-off-by: Scott M Stark <starksm64@gmail.com>

### DIFF
--- a/specification/src/main/asciidoc/platform/ApplicationAssemblyDeployment.adoc
+++ b/specification/src/main/asciidoc/platform/ApplicationAssemblyDeployment.adoc
@@ -1211,6 +1211,17 @@ descriptor but do not appear in the application deployment descriptor,
 the deployment tool must use the description provided in the module
 deployment descriptor.
 
+[NOTE]
+====
+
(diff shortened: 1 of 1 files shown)


kazumura:
The description of "java" and "javax" may be redundant and may not be needed in this context.
There are a lot of restricted packages other than "java" and "javax".
If we would say about "java" and "javax" at the beginning of this paragraph,
we should also say that the behavior of an application which uses "java" and "javax" is undefined.

ivargrimstad:
> The description of "java" and "javax" may be redundant and may not be needed in this context. There are a lot of restricted packages other than "java" and "javax". If we would say about "java" and "javax" at the beginning of this paragraph, we should also say that the behavior of an application which uses "java" and "javax" is undefined.

I agree. We should limit the statement to the namespace that is within our control.

starksm64:
Updated to only discuss the jakarta package restriction

kazumura (approved):

ivargrimstad (approved):

PR #486, opened 2022-05-05
Refine module-info.class description to clarify they are not standard

dblevins:
Explicitly call out that the module-info.class files are non-standard.

There was conversation where everyone understood there were no testing requirements and vendors did not have to support them, but in the same statement some people also called the module-info.class files in the spec project API jars "standard", "official" and "spec defined."  We need to be explicit that the module-info.class files are none of those things.


starksm64 (approved):

dblevins:
Thank you, @starksm64 !!

PR #489, opened 2022-05-13
Add minutes for May 10 Platform call

ivargrimstad:


starksm64 (commented):
I was on the call, and did add my name, but when I went to check the minutes, it was not there.

ivargrimstad:
> I was on the call, and did add my name, but when I went to check the minutes, it was not there.

No worries, I'll add you

PR #609, opened 2023-02-01
Refine proposed draft after TCK Project input

ivargrimstad:


starksm64:
Red Hat has a number of TCK issues that we want resolved before an EE 11 release. The April 5 TCK meeting will start drilling into a roadmap, but we won't support a Q1 2024 release of EE 11. We believe the technical debt issues in the TCK must be a blocker for the EE 11 release.

PR #746, opened 2023-09-07
A prototype of adding persistence requirements for CDI in EE environments

starksm64:
This is a prototype of including component spec integration requirements in a platform specification. This is based on @gavinking persistence PR https://github.com/jakartaee/persistence/pull/460

Notes:
- The persistence_3_0.xsd file will not be here in the future. It is here now because the embedded complex type of the persistence-unit element needed to be pulled to the top level in order for it to be extended by the persistence-cdi-3.2.xsd schema
- The `jakarta.persistence.spi.PersistenceUnitInfo` interface is no longer a complete representation of the extended persistence.xml file. Maybe a jakarta.persistence.spi.cdi.PersistenceUnitExtInfo subinterface could be created to add the changes missing from the persistence PR.

Co-authored-by: @gavinking


specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,148 @@
...
+- the qualifiers specified by _qualifier_ XML elements in
+_persistence.xml_, or _jakarta.enterprise.inject.Default_,
+if no qualifiers are explicitly specified,
+- scope _jakarta.enterprise.context.Dependent_,
+- no interceptor bindings,
+- a bean implementation which simply obtains the instance of
+the bean type by calling the appropriate getter method of
+the _EntityManagerFactory_ bean.
+
+==== Obtaining an Entity Manager Factory using CDI
```
scottmarlow, line 44: This `Obtaining an Entity Manager Factory using CDI` section is a duplicate of same above.

gavinking: One of these passages is meant to be about injecting the `EntitiyManager`. Copy/paste error, I guess.

gavinking:
Thanks, @starksm64 🙏

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,148 @@
...
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- persistence.xml schema -->
+<xs:schema targetNamespace="https://jakarta.ee/xml/ns/persistence"
+            xmlns:xs="http://www.w3.org/2001/XMLSchema"
+            xmlns:persistence="https://jakarta.ee/xml/ns/persistence"
+            elementFormDefault="qualified"
+            attributeFormDefault="unqualified"
+            version="3.0">
+
+    <xs:include schemaLocation="persistence_3_0.xsd" />
```
lukasj, line 94: I am not a big fan of this way of extending schema - the drawback I see here is that the schema has to be updated every time the original one changes.

I'm thinking about an alternative - what if we add an extension point to the persistence schema itself somewhere? Sth like:

```
    <xsd:element name="persistence">
        <xsd:complexType>
            <xsd:sequence>
                <xsd:element name="persistence-unit" minOccurs="1" maxOccurs="unbounded">
                    <xsd:complexType>
                        <xsd:sequence>
                            <xsd:element name="description" type="xsd:string" minOccurs="0"></xsd:element>
                            <xsd:element name="provider" type="xsd:string" minOccurs="0"></xsd:element>
                            ...
                            <xsd:element name="properties" minOccurs="0">
                           <xsd:any namespace="##other" processContents="lax"/>
                        </xsd:sequence>
                    </xsd:complexType>
                </xsd:element>
            </xsd:sequence>
        </xsd:complexType>
    </xsd:element name="persistence">
...
```
the usage then could look like:
```
    <persistence-unit xsi:type="persistenceUnitExtType" name="my-unit" xmlns:cdi="https://jakarta.ee/xml/ns/persistence-cdi">
...
        <cdi:scope>com.example.jpa.ACustomScope</cdi:scope>
        <cdi:qualifier>com.example.jpa.CustomQualifier</cdi:qualifier>
    </persistence-unit>
</persistence>
```

The question is - does this _extension_ need to be available through `PersistenceUnitInfo`? If so, how? Ideally, it should be `org.w3c.dom.Node`, but I don't think I like the use of that type here

starksm64: If it does not need to be available via PersistenceUnitInfo, then there does not need for an extension to persistence.xml. In that case we could just require the info be passed into the Map passed to the factory method. Otherwise, your suggestion about the alternative syntax looks good to me.

gavinking:
```xml
  <persistence-unit name="my-unit" xmlns:cdi="https://jakarta.ee/xml/ns/persistence-cdi">
     ...
    <cdi:scope>com.example.jpa.ACustomScope</cdi:scope>
    <cdi:qualifier>com.example.jpa.CustomQualifier</cdi:qualifier>
  </persistence-unit>
````

This looks nice indeed.

**Review by scottmarlow — APPROVED** [review]
I am giving my +1 approval here and for https://github.com/jakartaee/persistence/pull/460.  I like the idea of the change regardless of which pull request is used.

**Comment by scottmarlow** [comment]
I think that a Jakarta EE implementation could use CDI [jakarta.enterprise.inject.spi.AfterDeploymentValidation](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#after_deployment_validation) event to know when the BeanManager becomes available for (CDI enabled) application deployments.  WildFly is already [using AfterDeploymentValidation](https://github.com/wildfly/wildfly/blob/main/jpa/subsystem/src/main/java/org/jboss/as/jpa/beanmanager/BeanManagerAfterDeploymentValidation.java) to help with our custom two-phase approach of Persistence Unit deployment bootstrapping. 

One of the parts that I am missing is how WildFly will be notified to call the EntityManager.close() for an entity manager for each of the different CDI scopes (assuming we do this specifically to work with CDI scopes)?

**Comment by bstansberry** [comment]
@scottmarlow Does https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#declaring_disposer_method illustrate one way this could be done?

**Comment by scottmarlow** [comment]
> @scottmarlow Does https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#declaring_disposer_method illustrate one way this could be done?

Yes, that does help I think, still reading for more context!  Thanks Brian!

**Comment by bstansberry** [comment]
I mentioned on this week's EE platform call how if integration requirements between two specs are covered in the platform spec instead of in the individual specs, then we need to think about what that means in the context of a modern appserver where ideally the user is only provisioning a subset of the platform. I'd only glanced at this at the time, and I see now it looks like you were thinking about that, @starksm64, i.e. by putting this in a section scoped to Web Profile, and referring to a "Jakarta EE environment".  That's good, particularly the latter.

I think we should try and flesh out a bit what we mean by a Jakarta EE environment in the current day and age. For example give a facelift to section 2 of the platform spec. And then the spec can discuss these kinds of requirements with reference to what we say there, instead with reference to profiles.

I don't think what I'm saying needs to block this; currently profiles are the finest level of granularity above the individual specs, so scoping requirements to them is what's possible now. I'm mentioning this because this is an example of the kind of disconnect there is between how the EE spec works and how modern implementations work.

File: specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc [hunk 3]
```diff
@@ -0,0 +1,148 @@
...
+- no interceptor bindings,
+- a bean implementation which simply obtains the instance of
+the bean type by calling the appropriate getter method of
+the _EntityManagerFactory_ bean.
+
+==== Obtaining an Entity Manager Factory using CDI
+
+A container environment might feature built-in integration of
+Jakarta Persistence with the CDI bean manager, allowing
+injection of a container-managed entity manager factory using
```
scottmarlow, line 48: Instead of referring to a `container-managed entity manager factory`, we might instead focus on the `container-managed entity manager` or even a `CDI-managed entity manager`.  

It is also 100% accurate to also talk about the container-managed entity manager factory because that is what the EE implementations/runtimes provide which is minimally obtained from the https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/spi/persistenceprovider#createContainerEntityManagerFactory(jakarta.persistence.spi.PersistenceUnitInfo,java.util.Map) call, however that isn't a term defined by the Jakarta Persistence specification (e.g. applications using `@PersistenceUnit EntityManagerFactory emf;`.

scottmarlow:
One (new?) term that I think came up during this week's platform call is `CDI Managed Persistence Contexts`  to describe what applications could use.  If anyone comes up with a more descriptive term, we should consider that.  Unless we decide to make `CDI Managed Persistence Contexts` to follow the same rules as `Container Managed Persistence Contexts`.

Currently the Persistence Specification defines the rules for how `[Container Managed Persistence Contexts](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#container-managed-persistence-contexts)` (also known as `container-managed entity managers`) work in general.  Which includes that applications calling the EntityManager.close() will see a IllegalStateException thrown if the entity manager is container-managed (no problem if the entity manager is `application-managed`.)  Also, there are specific rules about the container-managed entity manager joining active (Jakarta Transactions) transactions and the different situations + expectations.

scottmarlow:
> I mentioned on this week's EE platform call how if integration requirements between two specs are covered in the platform spec instead of in the individual specs, then we need to think about what that means in the context of a modern appserver where ideally the user is only provisioning a subset of the platform. I'd only glanced at this at the time, and I see now it looks like you were thinking about that, @starksm64, i.e. by putting this in a section scoped to Web Profile, and referring to a "Jakarta EE environment". That's good, particularly the latter.
> 
> I think we should try and flesh out a bit what we mean by a Jakarta EE environment in the current day and age. For example give a facelift to section 2 of the platform spec. And then the spec can discuss these kinds of requirements with reference to what we say there, instead with reference to profiles.
> 
> I don't think what I'm saying needs to block this; currently profiles are the finest level of granularity above the individual specs, so scoping requirements to them is what's possible now. I'm mentioning this because this is an example of the kind of disconnect there is between how the EE spec works and how modern implementations work.

Thanks Brian, I think these are great thoughts and since we are introducing a new thing and we aren't restricted to previous rules.  

If we do target `Web Profile` does that mean `CDI Managed Persistence Contexts` would only work at a minimal with required `Web Profile` technologies?  For example, `Jakarta Transactions` is a required technology so that would be expected to be present in the EE server.

Perhaps some EE implementations could deploy `CDI Managed Persistence Context` applications without `Jakarta Transactions` being on the application classpath meaning that deployed applications would not use `Jakarta Transactions` at all.  These applications would strictly not use XA transactions which means they are limited to one database schema (e.g. database user login account used) per application database transaction.  So IMO it might also be interesting to target less than the `Web Profile` as well.

scottmarlow:
Regardless of the target EE profile, I would like to point out the impact on compatibility between application use of `CDI Managed Persistence Context` and `Container Managed Persistence Context` so we can all understand what will work and what won't.  I'm going to update this comment with the details in a bit...

bstansberry:
> >
> > I don't think what I'm saying needs to block this; currently profiles are the finest level of granularity above the individual specs, so scoping requirements to them is what's possible now. I'm mentioning this because this is an example of the kind of disconnect there is between how the EE spec works and how modern implementations work.
> 
> Thanks Brian, I think these are great thoughts and since we are introducing a new thing and we aren't restricted to previous rules.
> 
> If we do target `Web Profile` does that mean `CDI Managed Persistence Contexts` would only work at a minimal with required `Web Profile` technologies? For example, `Jakarta Transactions` is a required technology so that would be expected to be present in the EE server.
> 
> Perhaps some EE implementations could deploy `CDI Managed Persistence Context` applications without `Jakarta Transactions` being on the application classpath meaning that deployed applications would not use `Jakarta Transactions` at all. These applications would strictly not use XA transactions which means they are limited to one database schema (e.g. database user login account used) per application database transaction. So IMO it might also be interesting to target less than the `Web Profile` as well.

@scottmarlow This is exactly the kind of question I'm getting at. But, I think the thing to do is for me to file a separate issue re this broader topic and discussions can move there. Now that I've raised the general point I don't want to overly derail the specific discussion about this particular change.

starksm64:
So to date the Jakarta TCKs and platform specs only deal with the existence
of the specs that make up the profile or platform or a greater superset.
Subsetting is strictly undefined. For that level of details there would
need to be a decision tree on what integration requirements existed at each
node where a new specification was added.

bstansberry:
The big issue with what I'm getting at is TCKs, and as a practical matter that problem is likely a showstopper for doing something different from what you've done with this kind of thing.

In terms of spec language it should be possible to do something to define a "Jakarta EE environment". Section 2.2 of the spec gives a nod at that. In theory even the Core Profile could be defined as the minimum base.

But testing subsetting in a TCK seems impractical. And if it's not tested, having it covered in spec language doesn't mean it's truly portable. Someone provisioning a subset is still just relying on their vendor's guarantee that it works.

scottmarlow:
> But testing subsetting in a TCK seems impractical. And if it's not tested, having it covered in spec language doesn't mean it's truly portable. Someone provisioning a subset is still just relying on their vendor's guarantee that it works.

The Platform TCK signature testing does validate that no sub-setting or super-setting is done by an implementation that is being tested.  There are platform-tck mailing list discussions from EE-9 time frame that capture feedback from a number of people that some express (long term held) concerns about forking EE in the way that Unix was forked years ago (and the damage that could cause to EE).  Others expressed feedback that it would be better to adjust TCK Signature testing to only validate (Java) binary compatibility to free implementations to do more.  As you said Brian, there will be a new issue for the broader discussion.  I just wanted to echo this little bit of info so we (or I can remember to) later link to that discussion about subsetting + supersetting on the other (broader) issue to be created.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
hantsy: CDI-managed EMF is `app managed`?

scottmarlow: @hantsy EMF is just the EMF, entity managers created by the application calling [EntityManagerFactory#createEntityManager](https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/entitymanagerfactory#createEntityManager()) are always app managed entity managers.

I'd like to list possible design choices [here](https://github.com/jakartaee/platform/pull/746#issuecomment-1812727696) for discussion regarding what exactly an application could do with a CDI-Managed entity manager (persistence context) along the lines of whether the underlying entity manager that is used for a CDI-Managed entity manager can be shared with `@PersistenceContext EntityManager myentitymanager;`.

starksm64:
The copy/paste error has been corrected and the configuration of the CDI scope/qualifier is based on the extension point as proposed in https://github.com/jakartaee/persistence/pull/566 that @lukasj suggested.

edburns:
Briefly reviewed this on 2024-02-27. We'll not include this in M2 at this point.

gavinking:
Folks, since the merge of https://github.com/jakartaee/persistence/pull/460, Jakarta Persistence now has a well-defined way to assign a CDI  `@ScopeType` and `@QualifierType`s to a persistence unit, via the `persistence.xml` document.  (Actually, since those annotations were moved from CDI to `jakarta.inject` many years ago, this actually works for _any_ dependency injection container, not just a CDI bean container.)

However, it's going to be up to the platform to define what these annotations mean in the Jakarta EE platform.

Here's what I think it should ideally say:

<blockquote>
<h3>Obtaining an EntityManager or EntityManagerFactory using CDI</h3>

 In the Jakarta EE environment, the container must feature built-integration of Jakarta Persistence
with the CDI bean manager, allowing injection of a container-managed entity manager using the annotation _jakarta.inject.Inject_.

For each persistence unit, the container  must make available a bean with:

- bean type `EntityManager`,
- the qualifiers specified by `qualifier` XML elements in `persistence.xml`, or `jakarta.enterprise.inject.Default`, if no qualifiers are explicitly specified,
- the scope specified by the `scope XML element in `persistence.xml`, or `jakarta.transaction.TransactionScoped` if no scope is explicitly specified,
- no interceptor bindings,
- a bean implementation which satisfies the requirements for a container-managed entity manager, as defined by the Jakarta Persistence specification.

Furthermore, the must make available a bean with:

- bean type `EntityManagerFactory`,
- the qualifiers specified by `qualifier` XML elements in `persistence.xml`, or `jakarta.enterprise.inject.Default`, if no qualifiers are explicitly specified,
- scope `jakarta.enterprise.context.ApplicationScoped`,
- bean name given by the name of the persistence unit,
- no interceptor bindings,
- a bean implementation which satisfies the requirements for a container-managed entity manager factory, as defined by the Jakarta Persistence specification.

Furthermore, the container must make available five beans with:

- bean types `CriteriaBuilder`, `PersistenceUnitUtil`, `Cache`, `SchemaManager`, and `Metamodel`, respectively,
- the qualifiers specified by `qualifier` XML elements in `persistence.xml`, or `jakarta.enterprise.inject.Default`, if no qualifiers are explicitly specified,
- scope `jakarta.enterprise.context.Dependent`,
- no interceptor bindings,
- a bean implementation which simply obtains the instance of the bean type by calling the appropriate getter method of the `EntityManagerFactory` bean.
<blockquote>

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+=== Jakarta Persistence & Jakarta Context and Dependency Injection (CDI) Integration
+This section specifies the requirements for Jakarta EE environments that integrate both {jpa-spec}[Persistence] and {cdi-spec}[CDI].
+
+==== Obtaining an Entity Manager using CDI
+
+A container environment might feature built-in integration of
+Jakarta Persistence with the CDI bean manager, allowing injection
+of a container-managed entity manager using the annotation
+_jakarta.inject.Inject_. In the Jakarta EE environment, the
+container must feature built-integration of Jakarta Persistence
```
bstansberry, line 13: s/built-integration/built-in integration/g

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+- a bean implementation which satisfies the requirements of
+this specification for a container-managed entity manager.
+
+==== Obtaining an Entity Manager Factory using CDI
+
+A container environment might feature built-in integration of
+Jakarta Persistence with the CDI bean manager, allowing
+injection of a container-managed entity manager factory using
+the annotation _jakarta.inject.Inject_. In the Jakarta EE
+environment, the container must feature built-integration of
```
bstansberry, line 36: s/built-integration/built-in integration/g

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+_persistence.xml_, or _jakarta.enterprise.inject.Default_,
+if no qualifiers are explicitly specified,
+- scope _jakarta.enterprise.context.Dependent_,
+- no interceptor bindings,
+- a bean implementation which simply obtains the instance of
+the bean type by calling the appropriate getter method of
+the _EntityManagerFactory_ bean.
+
+==== persistence.xml Extended Schema
+The Persistence 3.1 specification schema includes an extension point that allows one to add elements for configuration of integration concerns. The required approach for declaring a CDI  scope and qualifier elements.
```
bstansberry, line 67: "The required approach.... elements." is not a complete sentence.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+      </properties>
+      <!-- CDI integration -->
+      <cdi:scope>com.example.jpa.ACustomScope</cdi:scope>
+      <cdi:qualifier>com.example.jpa.CustomQualifier</cdi:qualifier>
+  </persistence-unit>
+
+</persistence>
+----
+
+====  Additional EntityManagerFactory Properties
```
bstansberry, line 97: should this be styled? i.e. _EntityManagerFactory_

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+The following additional properties correspond to the
+new elements and properties in the _persistence.xml_ file when using the persistence-cdi-3.2xsd schema and associated persistenceUnitExtType. When any of these
```
bstansberry, line 100: _persistenceUnitExtType_

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+properties are specified in the Map parameter passed to the
```
bstansberry, line 101: _Map_

edburns:
Hello @starksm64 , this PR is still in Draft state. Is this ready for review?

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+built-in integration must make available a bean with:
+
+- bean type _EntityManager_,
+- the qualifiers specified by _qualifier_ XML elements in
+_persistence.xml_, or _jakarta.enterprise.inject.Default_,
+if no qualifiers are explicitly specified,
+- the scope specified by the _scope_ XML element in
+_persistence.xml_, or _jakarta.transaction.TransactionScoped_
+if no scope is explicitly specified,
+- no interceptor bindings,
```
arjantijms, line 26: What is the intend here? That just by default there are no interceptor bindings, or, that it is forbidden for interceptors to be defined for this type?

arjantijms:
@starksm64 @gavinking @lukasj 

Except for a small question about interceptor bindings this looks fine to me.

I'm just wondering, why couldn't we have this text in the Persistence spec in a chapter or section titled "Container / Platform responsibilities" and opening with something like "In a Jakarta EE environment the following additional requirements hold" (or similar words to that effect)?

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
jhanders34, line 67:
```suggestion
The Persistence 3.2 specification schema includes an extension point that allows one to add elements for configuration of integration concerns. The required approach for declaring a CDI  scope and qualifier elements.
```

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+[source,xml]
+----
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- persistence.xml -->
+<persistence xmlns="https://jakarta.ee/xml/ns/persistence"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://jakarta.ee/xml/ns/persistence
+          https://jakarta.ee/xml/ns/persistence/persistence_3_1.xsd"
+        version="3.1">
```
jhanders34, line 78:
```suggestion
          https://jakarta.ee/xml/ns/persistence/persistence_3_2.xsd"
        version="3.2">
```

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
Emily-Jiang: If we allow to supply a qualifier to avoid some migration concerns, we need an example on how to specify the qualifier in the `persistence.xml`. I guess the .xsd needs to be updated. Is this too late since Jakarta Persistence has been released. I think we should just say the `Default` qualifier will be used. I am not sure how you can add scope either in the `persistence.xml` unless it was done already.

Emily-Jiang: See the later section on the explanation. Please add a reference to the later section on the xsd.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
+:cdi-spec: https://jakarta.ee/specifications/cdi/4.1
+:jpa-spec: https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1
+
+=== Jakarta Persistence & Jakarta Context and Dependency Injection (CDI) Integration
+This section specifies the requirements for Jakarta EE environments that integrate both {jpa-spec}[Persistence] and {cdi-spec}[CDI].
+
```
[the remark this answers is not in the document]
scottmarlow: > I am confused for instance:

@Inject
EntityManager myEntityManager
>
> Which EntityManager is returned? The spec logic seems to imply that this should work and return an EntityManager even though scope and qualifiers were not provided. I think for EntityManagerFactory it is explained more for this case, but I don't know if it is for EntityManager unless I am just missing it.

Wouldn't this be the same as:

```
@PersistenceContext 
EntityManager myEntityManager
```

I think that ^ would just use the first persistence unit found in the application deployment by the Persistence container.  I don't recall any guidance in the Persistence spec on what what happens if there are multiple persistence units defined.

WildFly has a persistence unit hint for marking the default persistence unit (a deployment failure is thrown if there are multiple default persistence units).  I thought we might of discussed this years ago on the jpa spec mailing list but I'm not really sure.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
gavinking: > That just by default there are no interceptor bindings, or, that it is forbidden for interceptors to be defined for this type?


@arjantijms I'm not sure I understand what you mean by this question, because interceptor bindings belong to beans, not to bean _types_.

 Of course one could define a bean with type `Entity Manager ` and interceptor bindings, but the bean defined _here_ does not have any interceptor bindings, because we don't provide any place to specify such bindings. (We would need to add an element to `persistence.xml`.)

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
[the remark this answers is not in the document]
gavinking: > Which `EntityManager` is returned?

The interpretation of such an injection point is specified in great detail by the CDI spec. It's not the place of this text to specify how beans are matched to injection points. All we need to specify here is what are the bean types and qualifiers the bean defined here. Then CDI itself takes care of the rest (according to nontrivial rules).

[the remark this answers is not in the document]
gavinking: > Wouldn't this be the same as:
>
>     @PersistenceContext 
>     EntityManager myEntityManager
>
> I think that ^ would just use the first persistence unit found in the application deployment by the Persistence container. I don't recall any guidance in the Persistence spec on what what happens if there are multiple persistence units defined.

No. CDI-based injection is never based on "the first thing found". It is always an error if an injection point is ambiguous.

All this is covered in section 5.2 of the CDI 4.1 spec. See, especially, section 5.2.2.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+==== Obtaining an Entity Manager using CDI
+
+A container environment might feature built-in integration of
+Jakarta Persistence with the CDI bean manager, allowing injection
+of a container-managed entity manager using the annotation
+_jakarta.inject.Inject_. In the Jakarta EE environment, the
+container must feature built-integration of Jakarta Persistence
+with the CDI bean manager.
```
gavinking, line 14: I originally wrote this text for inclusion in the Persistence spec, but since it's now proposed for the platform spec, I believe it should read:

```suggestion
A Jakarta EE container must feature built-integration of Jakarta 
Persistence with the CDI bean manager, allowing injection
of a container-managed entity manager using the annotation
_jakarta.inject.Inject_.
```

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+this specification for a container-managed entity manager.
+
+==== Obtaining an Entity Manager Factory using CDI
+
+A container environment might feature built-in integration of
+Jakarta Persistence with the CDI bean manager, allowing
+injection of a container-managed entity manager factory using
+the annotation _jakarta.inject.Inject_. In the Jakarta EE
+environment, the container must feature built-integration of
+Jakarta Persistence with the CDI bean manager.
```
gavinking, line 37: Similarly:

```suggestion
A Jakarta EE container must feature built-integration of
Jakarta Persistence with the CDI bean manager, allowing
injection of a container-managed entity manager factory using
the annotation _jakarta.inject.Inject_.
```

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+A container environment might feature built-in integration of
+Jakarta Persistence with the CDI bean manager, allowing injection
+of a container-managed entity manager using the annotation
+_jakarta.inject.Inject_. In the Jakarta EE environment, the
+container must feature built-integration of Jakarta Persistence
+with the CDI bean manager.
+
+For each persistence unit, a container which features such
+built-in integration must make available a bean with:
```
gavinking, line 17:
```suggestion
For each persistence unit, the container must make available 
a bean with:
```

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+A container environment might feature built-in integration of
+Jakarta Persistence with the CDI bean manager, allowing
+injection of a container-managed entity manager factory using
+the annotation _jakarta.inject.Inject_. In the Jakarta EE
+environment, the container must feature built-integration of
+Jakarta Persistence with the CDI bean manager.
+
+For each persistence unit, a container which features such
+built-in integration must make available a bean with:
```
gavinking, line 40:
```suggestion
For each persistence unit, the container must make available 
a bean with:
```

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+- bean type _EntityManagerFactory_,
+- the qualifiers specified by _qualifier_ XML elements in
+_persistence.xml_, or _jakarta.enterprise.inject.Default_,
+if no qualifiers are explicitly specified,
+- scope _jakarta.enterprise.context.ApplicationScoped_,
+- bean name given by the name of the persistence unit,
+- no interceptor bindings,
+- a bean implementation which satisfies the requirements of
+this specification for a container-managed entity manager
+factory.
```
gavinking, line 51:
```suggestion
- a bean implementation which satisfies the requirements of
the Persistence specification for a container-managed entity 
manager factory.
```

gavinking (changes requested):

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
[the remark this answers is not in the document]
scottmarlow: Thanks, I will look at [CDI spec section 5.2.2](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#unsatisfied_and_ambig_dependencies).

The Persistence Specification rules that would apply to `container-managed persistence contexts` and how they resolve referenced persistence units is https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#a12459

The current thinking (from a brief discussion this week during the Jakarta EE Platform call) is that the [Persistence spec contract for container-managed persistence contexts](https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#container-managed-persistence-contexts) would be followed here as well which includes other Persistence Spec requirements specified through-out the Spec + Javadoc text.  

I think the ask for all of us is to figure out if this Platform change can be included in Jakarta EE 11.

I think that the https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#a12459 section does state how a particular persistence unit could be used if it is the "first thing found" as per `When referencing a persistence unit using the unitName annotation element or persistence-unit-name deployment descriptor element, the visibility scope of the persistence unit is determined by its point of definition` part of the `8.2.2. Persistence Unit Scope` spec text.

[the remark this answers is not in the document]
gavinking: Look, I think we need to take a step back here and understand that CDI is very powerful and that most of the things we need here are already defined by CDI and we don't need to tie ourselves in knots re-saying things that CDI already says. Sure, for historical reasons the Persistence spec says a bunch of stuff that defines how the legacy `@PersistenceUnit` and `@PersistenceContext` annotations work, but we're trying to move away from that: people have asked for _CDI integration_ and that means defining things in terms of CDI concepts.

The proposed text is:

> the container must make available a bean with ... the scope specified by the _scope_ XML element in _persistence.xml_, or _jakarta.transaction.TransactionScoped_ if no scope is explicitly specified

The semantics of the `@TransactionScoped` annotation already tell us everything we need to know about the default lifecycle of the persistence context: and, yes, the persistence context management is going to be essentially the same as what is defined by section 7.7.2 _Container-managed Transaction-scoped Persistence Context_ of the Persistence spec, but instead of being some special thing connected to the `@PersistenceContext` annotation, it's just something we naturally get for free from CDI. Alternatively, if you explicitly specify a _different_ scope in `persistence.xml`, then CDI controls the lifecycle according to the semantics of that scope.

Similarly, this text:

> the container must make available a bean with bean type _EntityManager_, the qualifiers specified by _qualifier_ XML elements in _persistence.xml_, or _jakarta.enterprise.inject.Default_, if no qualifiers are explicitly specified

completely determines how the persistence unit is matched to an injection point, _according to the rules of CDI_. We don't need to look anywhere else than in the CDI spec to find how that works.

The point is that nobody should be looking in the Persistence spec to find out how CDI injection works, or about the lifecycle of contextual beans in CDI.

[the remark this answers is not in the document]
gavinking: P.S. An example of how **NOT** to do this stuff is provided by sections 12.4.3 and 12.4.4 of the current JMS spec, which purports to define injection of the `JMSContext` via CDI, but doesn't use the language/concepts of CDI.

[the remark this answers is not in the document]
gavinking: Oh, perhaps confusion arises from the use of the phrase "container-managed entity manager". This is meant in the generic sense of an entity manager which is created and destroyed by the container. It's not a reference to section 7.7 of the Persistence spec. (Though of course the semantics are *extremely* similar by default.)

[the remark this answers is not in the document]
scottmarlow: > The semantics of the @TransactionScoped annotation already tell us everything we need to know about the default lifecycle of the persistence context: and, yes, the persistence context management is going to be essentially the same as what is defined by section 7.7.2 Container-managed Transaction-scoped Persistence Context of the Persistence spec, but instead of being some special thing connected to the @PersistenceContext annotation, it's just something we naturally get for free from CDI. Alternatively, if you explicitly specify a different scope in persistence.xml, then CDI controls the lifecycle according to the semantics of that scope.

At a concrete level, I get that EE deployer code that is already deploying the persistence unit via a call to the PersistenceProvider#createContainerEntityManagerFactory(jakarta.persistence.spi.PersistenceUnitInfo) will meet the additional requirements for CDI in EE environments in a way that integrates with non-CDI code (for the container-managed + transaction-scoped case).  However, I must be wrong as you said that the EE deployer code isn't finding the right persistence unit to use, as that is something we get for free from CDI.

Again, at a concrete level, how does CDI get access to all of the different deployed persistence units that only the EE deployer (internals) has access to?  

Is that made possible by the EE deployer adding something like a https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#declaring_disposer_method (perhaps after jakarta.enterprise.inject.spi.AfterDeploymentValidation is fired?)  

I could see how the deployer_method example linked above could have access to the deployed persistence unit.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
lukasj: does the explicit usage of the spec version number make sense here? The version _in action_ is mentioned somewhere at the beginning already

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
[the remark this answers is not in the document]
gavinking: > However, I must be wrong as you said that the EE deployer code isn't finding the right persistence unit to use, as that is something we get for free from CDI.

CDI doesn't need to know anything about the persistence unit or how it is deployed. All CDI needs is for the container to hand it the definition of a bean.

> Again, at a concrete level, how does CDI get access to all of the different deployed persistence units that only the EE deployer (internals) has access to?

I mean to me this is an implementation detail of the bean that the EE container provides to CDI. We don't need to spell out how this works.

> Is that made possible by the EE deployer adding something like a https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#declaring_disposer_method (perhaps after jakarta.enterprise.inject.spi.AfterDeploymentValidation is fired?)

Yes, exactly, something like that.

It could be:

- a producer method (which creates an `EntityManager`) + disposer method (which calls `close()` on it), 
- a bean definition registered by a portable extension, or even
- something more implementation-specific in the case when the CDI bean manager is integrated more deeply with the rest of the EE container via vendor-defined SPIs.

Of course the Persistence spec still defines how the container interacts with the Persistence provider in order to bootstrap a persistence unit and create an `EntityManagerFactory`. Nothing about that process changes.

The only new thing is that after creating the `EntityManagerFactory`, the container then makes certain beans available to CDI (most likely via some sort of portable extension, but, again, we don't need to specify that).

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+Furthermore, the container must make available five beans with:
+
+- bean types _CriteriaBuilder_, _PersistenceUnitUtil_, _Cache_,
+_SchemaManager_, and _Metamodel_, respectively,
```
lukasj, line 56: should these be made available for `@Resource` injection and/or DD override (=reachable from JNDI) and as such also some description added to chapter 5?

gavinking: @lukasj anything like that would require mechanisms external to CDI and would significantly complicate this whole thing.

gavinking: (And CDI itself already provides ways to override beans.)

lukasj: If this cannot be made consistent with the rest of resources available through the platform, I would not add this requirement for now

scottmarlow: > If this cannot be made consistent with the rest of resources available through the platform, I would not add this requirement for now

Or we could document in the Platform Specification what (Jakarta EE 11) portable applications can expect in the Platform Specification and what this new feature is not integrated with (e.g. Resource and/or DD override, I think extended persistence contexts but not completely sure about that).  

Having said that, I think we do need to prototype this requirement in EE implementations so we can see what EE features do not work (for EE implementations) that should work.

Based on this pr conversation, I understand more of what is expected but I think we need to see this feature actually being used so we can all see what happens when very large applications:
- use this feature with active JTA transaction + existing application code also using the same persistence contexts.
- use this feature with  with no active JTA transaction + existing application code.  

Also, the https://jakarta.ee/xml/ns/persistence/persistence_3_2.xsd already has the support for this feature and if we don't define how the Platform uses the new fields, we might end up with a lot different extensions from  the EE vendors, non of which are completely consistent.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
arjantijms: > > That just by default there are no interceptor bindings, or, that it is forbidden for interceptors to be defined for this type?
> 
> @arjantijms I'm not sure I understand what you mean by this question, because interceptor bindings belong to beans, not to bean _types_.


Yes, though the text now says:

```
For each persistence unit, a container which features such built-in integration must make available a *bean* with:

no interceptor bindings,
```

I would expect the spec to say what IS required. For things that are NOT required, I would expect those things to be either left out or explicitly saying they are not required.

The way it's written now would probably trigger a TCK engineer to write a test that very specifically checks in whatever way is possible that the injected EntityManager does absolutey not have any interceptor bindings.

If my implementation would now for some reason add interceptor bindings for internal use, say for tracing, or for transaction management or whatever, then this would not be allowed, seemingly. But then the question arises; WHY is this restriction in place?

But from your reply I taste you dont want to specificallty constraint the Bean to not have interceptor bindings, but just that by default it does not need any (spec defined) interceptor bindings.

>  because we don't provide any place to specify such bindings. 

Isn't that already available via an extension? I quite often programmatically add interceptor bindings to beans that way.

gavinking: > If my implementation would now for some reason add interceptor bindings for internal use, say for tracing, or for transaction management or whatever, then this would not be allowed, seemingly. But then the question arises; WHY is this restriction in place?

If the bean has interceptor bindings that are visible to the application developer, then that would result in non-portable behavior. Now, of course, an implementation could in principle have its own internal interceptor binding types, and the bean could in principle have such interceptor bindings. But then such binding types would not be visible to the application programmer, nor to a TCK engineer.

gavinking: > Isn't that already available via an extension? I quite often programmatically add interceptor bindings to beans that way.

I mean, that's fine, nothing in that language says that _portable extensions_ can't add interceptor bindings. Just the bean provided by the container does not have any.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
arjantijms: > should these be made available for @Resource injection and/or DD override (=reachable from JNDI)

I think it's time to let go of @Resource and JNDI. So from my point of view that would be 

-1 -> Do NOT make these available for @Resource injection and/or DD override (=reachable from JNDI)

gavinking: I agree. I don't see the value of `@Resource` or JNDI here. In my mind, that's legacy stuff that we've already moved away from a long time ago.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
arjantijms: > Just the bean provided by the container does not have any.

I hear you, though as a container I can use a portable extensions just as well of course.

The question is perhaps how it would result in non-portable behaviour. Say I would in GlassFish add an interceptor binding to the `Bean<T>` for `EntityManager#persist`,  to track how many times persist is called (I display a nice graph of this in the admin console).

Applications injecting `EntityManager` on my GF instace will have a working `persist` method, and that same code would work on WildFly just as well. Tracking persists calls is just an internal GF feature.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
lukasj: > > should these be made available for @resource injection and/or DD override (=reachable from JNDI)
> 
> I think it's time to let go of @resource and JNDI. So from my point of view that would be
> 
> -1 -> Do NOT make these available for @resource injection and/or DD override (=reachable from JNDI)

then I have to ask - why this functionality has not been deprecated yet? Shouldn't that be the very first step a platform should have taken in order to re-build itself on top of CDI?

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
gavinking: Well the application program or framework could associate an interceptor with the given binding type, and then that interceptor would be triggered in this container but not in other containers.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
arjantijms: >  why this functionality has not been deprecated yet? 

There's so many questions there. I think we should have done it in 2009 indeed, but by then the EJB group was still big and powerful, and I think we didn't want to aggitate them too much.

We also shouldn't have just stopped with @Transactional in Java EE 7, but did much more in that release.

It's a sad story alltogether how we coud reinvented ourselves many years ago, but somehow didn't, even though the path was clearly ahead of us.

Java EE/Jakarta Security was the first, in Java EE 8, to add all kinds of resources/services and complete base it on CDI. Completely and utterly ignoring @Resource/JNDI. People though it could not be done. Somehow everything would always need JNDI. But hey, Jakarta Security did it. And now Quarkus and MP also provide tons of services, and it's really true that JNDI is not needed.

Hopefully for EE 12 we can finally deprecate our old friend JNDI and co.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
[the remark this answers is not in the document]
arjantijms: > Again, at a concrete level, how does CDI get access to all of the different deployed persistence units that only the EE deployer (internals) has access to?

Maybe in different wording; CDI does not go looking for deployed persistence units. Instead, whatever code creates a persistence unit would register it with CDI under the scope and qualifiers associated with that persistence units.

The lookup is just normal CDI then.

From a cursory look I would say though that

```
@PersistenceContext 
EntityManager myEntityManager
```

and

```
@Inject
EntityManager myEntityManager
```

should ultimately result in the same entity manager, but I've got to study things a bit more to say for sure.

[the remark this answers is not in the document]
gavinking: <blockquote>
From a cursory look I would say though that

```
@PersistenceContext 
EntityManager myEntityManager
```

and

```
@Inject
EntityManager myEntityManager
```
should ultimately result in the same entity manager, but I've got to study things a bit more to say for sure.
</blockquote>

Mmmmm, the current wording doesn't guarantee that (though it also doesn't completely disallow it).

One the reason being: `@PrsistenceContext` always injects a transaction-scoped PC. Whereas the new functionality specified here allows PCs with arbitrary CDI scope, as specified in `persistence.xml`.

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
starksm64: Changed to Injecting...

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
starksm64: We have an issue that was delayed for EE10 on the deprecation of resource injection and JNDI.
https://github.com/jakartaee/platform/issues/355

specification/src/main/asciidoc/webprofile/jpa-cdi/CDI-JPA.adoc
starksm64: I have removed it from the text, but it does still show up in the schema used in the persistence.xml example.

bstansberry (approved):

starksm64:
@edburns from the https://jakarta.ee/committees/specification/versioning/ process, we need to make a statement on the deprecation of `@Resource` and `@Resources` on the common-annotations feature, ideally with `@Deprecation` on the API/javadoc which probably requires a minor release, and in the platform spec. JNDI is just a platform feature so it just needs to be deprecated there.


## Deprecation
API methods that are marked for deprecation should be annotated with @Deprecated, and where possible the JavaDoc for the method should be updated to point users towards the replacement (if there is one). This mandates a minor version increment.

As noted above, the “deprecation” of behaviour by changing it to something else can happen in two ways:
- Using a flag to switch from the old to the new behaviour in a minor release. The flag and support for the old behaviour can be dropped in a later major release
- Without using a flag to switch between the old and new behaviour in a major release. This should be avoided where possible to avoid sudden breaking changes.

Component features must be “deprecated” via a process similar to the Pruning process that existed before. A mandatory Component Feature is transitioned to being an optional feature in release N of the Component specification, and the optional feature should be marked as “Deprecated” in some manner (for example in a “Proposed for Removal” section of the specification). The optional feature may then be removed from the Component specification in release N+1 (or beyond). As noted above, the transition of the feature to being optional would mandate a major release, and the removal of it would mandate a subsequent major release.

An existing optional feature which has not yet been marked for deprecation must be marked as deprecated in a minor or major release, to then be removed in a subsequent major release.

It should be stressed that optional features do not have to be removed – the Component specification can maintain them as optional for as long as desired.

See the Platform Feature section for more details on how Component Features which are pruned in this manner are dealt with there.

Deprecation from a Platform specification stance has two scopes:

- Removal of Component specifications
- Removal of Component features

### Component Features
When a Component feature is transitioned from being mandatory to optional, the Platform specification and its profiles must include a version of the specification which maintains this change for at least one release. Before an optional feature can be removed from the Platform specification, it must be marked for at least one release under a “Proposed for Removal” section. When removing an optional feature from the Platform specification or its profiles a reference must be provided to the version of the specification that marked the feature as “Proposed for Removal”.

It should be noted that the Platform specification and its profiles are not allowed to “jump” a release to introduce breaking changes without applying for an exception This is to discourage a component specification quickly doing two or more back-to-back releases to quickly remove a feature.

The intent is that there must be a release of the Platform specification which contains the feature in a “Proposed for Removal” state before it can be removed, but this should not necessarily prevent a component specification from moving at a quicker pace than that of the Platform.

For an example scenario:

- A Component specification defines a mandatory feature in release 2.0 and this is included in the Platform specification release X.
- The Component specification transitions the feature to being an optional one in release 3.0 and marks it for removal.
- The Component specification removes the feature entirely in release 4.0.
- Platform specification release X+1 must contain Component specification 3.0, mark the optional Component feature as included, and can also mark the optional Component feature under the “Proposed for Removal” section – it is not allowed to immediately include Component specification 4.0 since this would bypass the “Proposed for Removal” step.
- Platform specification release X+2 can proceed to include version 4.0 of the Component specification with the feature removed. Alternatively, the Platform specification release X+2 may choose to maintain Component specification 3.0, but would not have to include the optional feature if it did not want to.

edburns:
See #962

PR #797, opened 2023-11-23
Apply changes in activation/mail 2.1 to the platform

lukasj:
EE 10 contains Activation/Mail 2.1 but it is not reflected there. This fixes version numbers of these specs and their names.

Also note that since there is no longer anything like default _JavaMail_ providing support for a set of protocols, the language in this part was changed to *should*, even though it may better to have it as a *must* - opinions?


ivargrimstad (approved):

scottmarlow:
Thanks @lukasj and thanks @ivargrimstad for reviewing!

PR #822, opened 2024-01-15
On branch edburns-msft-jea-283-revise-release-plan-gh-platform-820 Apply wording from release plan email.

### LINKED ISSUE #820: Jakarta EE 11 must support certification under Java SE 17 in addition to Java SE 21
**Is your feature request related to a problem? Please describe.**
The current Java SE 21 mandate is overreaching in terms of a requirement if EE 11 implementations cannot even boot using the non-normative API component specification artifacts under Java SE 17.

**Describe the solution you'd like**
While Java SE 21 can be a mandated runtime for certification, there still must be support for Java SE 17 as a valid runtime for certification.

**Describe alternatives you've considered**
We will have to fork Jakarta EE 11 specification projects that compile with a target language level of Java SE 21 to either produce a lower target version or a MR version of the artifact. If a TCK requires a Java SE 21 runtime it is unclear what approach can be taken. This will impact our ability to provide CIs for Jakarta Validation and Jakarta CDI.

**Additional context**

The following New Relic survey of Java version adoption is consistent with our user base adoption profile:
https://newrelic.com/resources/report/2023-state-of-the-java-ecosystem

image

Java SE 21 is not even in the picture because of its Q4 2023 release date. The survey shows that most adoption is skewed toward the current LTS(-1) version. A best case scenario is that Java SE 21 sees a large first year adoption rate, but our expectation is that similar to this survey, this still leaves more than 90% of our userbase still be on an LTS release prior to Java SE 21.

edburns:
Fixes #820.

modified:   jakartaee11/JakartaEE11ReleasePlan.md

- See [mail archive](https://www.eclipse.org/lists/jakartaee-platform-dev/msg04374.html).

### DIFF
--- a/jakartaee11/JakartaEE11ReleasePlan.md
+++ b/jakartaee11/JakartaEE11ReleasePlan.md
@@ -49,9 +49,9 @@ Continue to make CDI the single component model used across all of EE by removin
 
 With the help of the platform project, the CDI project will do the work to move all aspects of CDI that deal with integrating CDI with other specifications out of the CDI spec and into the platform spec or appropriate profile spec. The remaining content will still be called CDI.  For more information see the [CDI issue tracker](https://github.com/jakartaee/cdi/issues/687#issuecomment-1667009015).
 
-### API Source and Target Level
+### API Source Code `--release` Level
 
-For the Jakarta EE Platform, the `source` and `target` (or maven `release` property) level is 17. For the component specs, the Jakarta EE Platform requires the `source` and `target` (or maven `release` property) is at most 17, but component specifications can decide on a lower level.
+For the Jakarta EE Platform, the Java compiler `--release` option is 17. For the component specs, the Jakarta EE Platform requires the Java compiler `--release` option is **at most** 17, but component specifications can decide on a lower level.
 
 ### TCK Source Level
 
--- a/jakartaee11/JakartaEE11ReleasePlan.md
+++ b/jakartaee11/JakartaEE11ReleasePlan.md
@@ -49,6 +49,10 @@ Continue to make CDI the single component model used across all of EE by removin
 
 With the help of the platform project, the CDI project will do the work to move all aspects of CDI that deal with integrating CDI with other specifications out of the CDI spec and into the platform spec or appropriate profile spec. The remaining content will still be called CDI.  For more information see the [CDI issue tracker](https://github.com/jakartaee/cdi/issues/687#issuecomment-1667009015).
 
+### API Source and Target Level
+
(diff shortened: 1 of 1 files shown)


edburns:
Linked to [JEA-284](https://dev.azure.com/jakarta-ee-azdo/jakarta-ee-azdo/_workitems/edit/284).

Emily-Jiang (approved):

jamezp:
The wording "source level" is used in several places. I don't think that's bad by any means, but is there a place where we can define something like:

> Source level is defined as the maximum level of the byte code the API can run on.

Possibly something better than that TBH :) But just to make it clear that `--release 17` must be used. IMO we don't want to close the door on an API using a Multi-Release JAR.

Emily-Jiang (commented):
LGTM

jakartaee11/JakartaEE11ReleasePlan.md
```diff
@@ -48,20 +49,27 @@ Continue to make CDI the single component model used across all of EE by removin
...
+   - These need not be the same implementation. There can be one implementation that passes on 17 and a different one that passes on 21.
 
 #### Signature Tests
 In Jakarta EE 9.1, the framework for performing API Signature tests was updated to make it easier to test Signatures across multiple Java levels.
 The framework and/or API tests may need to be tweaked as we continue to learn and experiment with future levels of Java.
 
 ### Compatible Implementation (CI) Source Level
-How a Compatible Implementation supports the Java SE 21 runtime (or above) will be left as a vendor-defined solution.
+
+The long-standing policy of considering specification API binaries, in Maven Central or anywhere else, as a non-normative convenience remains unchanged. The platform project is silent on this matter. But because the platform project does mandate a specific JDK requirement for compatible implementations passing the component or platform TCK, specification binaries are practically constrained to follow that mandate. See [TCK Source Level](#tck-source-level)
```
bstansberry, line 72: I'm not clear on what 'follow that mandate' means here.  Is the idea that the "API Source Code --release Level" section specifies source code requirements, so here we're saying that when the spec publishes a binary it shouldn't somehow work around those requirements and build a binary at a binary level above 17?

bstansberry: I also have a mental hurdle moving from the first 2 sentences to the last too. I think the first too are meant as a kind of qualifier to say the last two don't are not to be read as creating a requirement to publish to Maven Central etc.

Perhaps the 3rd sentence can be:

"However, because the platform project does mandate a specific JDK requirement for compatible implementations passing the component or platform TCK, if a component specification does publish an API binary, that binary is practically constrained to follow that mandate."

edburns: Fixed in 7f264e0 . Thanks @bstansberry .

starksm64:
LGTM

PR #852, opened 2024-02-26
Indicate that JAXB and JAXWS do get version updates only.

edburns:


jakartaee11/JakartaEE11ReleasePlan.md
```diff
@@ -223,7 +223,7 @@ We are proposing to deliver Jakarta EE 11 in a set of waves similar to those del
 - Jakarta JSON Binding
 - Jakarta Mail* (service release)
 - Jakarta SOAP with Attachments
-- Jakarta XML Binding
+- Jakarta XML Binding* (version increment only)
```
Emily-Jiang, line 226: Can we update this to be (service release) so that it is consistent with others. @edburns What does `version increment only` means?

PR #857, opened 2024-03-08
On branch edburns-msft-jea-341-xml-schema-version-update-policy Apply suggestion from Emily.

edburns:
modified:   Versioning.md

I think this info should be documented in this [file](https://jakartaee.github.io/platform/Versioning) and then we should follow this rule in the future releases.

See [AB#341](https://dev.azure.com/jakarta-ee-azdo/7cf9a9d3-9854-41e0-9a66-99506a9ed874/_workitems/edit/341).

### DIFF
--- a/Versioning.md
+++ b/Versioning.md
@@ -65,6 +65,5 @@ Examples:
 We haven't been consistent in whether the first version of a product-specific
 descriptor should include a revision of "-0" or just omit the revision number.
 
-If a DTD or XSD file has normative changes during an EE release cycle, its version number and file name must be updated to match the version number of the component specification for that EE release.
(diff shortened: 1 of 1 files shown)


Versioning.md
```diff
@@ -64,3 +64,7 @@ Examples:
 
 We haven't been consistent in whether the first version of a product-specific
 descriptor should include a revision of "-0" or just omit the revision number.
+
+If a DTD or XSD file has normative changes during an EE release cycle, its version number and file name must be updated to match the version number of the component specification for that EE release.
```
Emily-Jiang, line 68: Since we will have to update the DTD or XSD file no matter whether there are changes or not, I would simplify the 2 paragraph with the following:


When a component specification is updated, the version number and file name of the corresponding XSD or DTD file must be updated to match the version number of the component specification for that EE release. The file content can be updated if applicable.

PR #1246, opened 2026-03-27
improve wording around context initialization/destruction events

Ladicek:
The CDI EE part of https://github.com/jakartaee/cdi/pull/958

### DIFF
--- a/specification/src/main/asciidoc/platform/cdi-ee-spec/scopescontext_ee.adoc
+++ b/specification/src/main/asciidoc/platform/cdi-ee-spec/scopescontext_ee.adoc
@@ -116,9 +116,9 @@ The session scope is active:
 The session context is shared between all servlet requests that occur in the same HTTP session.
 The session context is destroyed when the `HTTPSession` times out, after all `HttpSessionListener` s have been called, and at the very end of any request in which `invalidate()` was called, after all filters and `ServletRequestListener` s have been called.
 
-An event with qualifier `@Initialized(SessionScoped.class)` is synchronously fired when the session context is initialized.
-An event with qualifier `@BeforeDestroyed(SessionScoped.class)` is synchronously fired when the session context is about to be destroyed, i.e. before the actual destruction.
-An event with qualifier `@Destroyed(SessionScoped.class)` is synchronously fired when the session context is destroyed, i.e. after the actual destruction.
+An event with qualifier `@Initialized(SessionScoped.class)` is synchronously fired immediately after the session context is initialized.
+An event with qualifier `@BeforeDestroyed(SessionScoped.class)` is synchronously fired immediately before the session context is destroyed.
+An event with qualifier `@Destroyed(SessionScoped.class)` is synchronously fired immediately after the session context is destroyed.
 The event payload is `jakarta.servlet.http.HttpSession`.
 
 [[application_context_ee]]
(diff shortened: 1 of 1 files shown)


arjantijms:
Just an idea, as this particular document is always maintained and reviewed by people from the CDI spec; why is it even in the platform?

Should we not, for consistency, move this document to the CDI spec itself as "Part III - CDI in Jakarta EE"?

We now have a "Part II - CDI in Java SE". Obviously that part isn't in the Java SE spec either.

Food for though?

arjantijms (approved):

Ladicek:
That's how it originally was. We entertained the idea of splitting the CDI spec into CDI and CDI EE, with two different lifecycles, all under the CDI project. For some reason, that didn't gain traction, but what _did_ gain traction was the idea of moving the CDI EE spec directly into Jakarta Platform. I'm not terribly happy about it, but if we were to move it back to CDI, the only sensible way IMHO would be to create a second spec under the CDI project.

manovotn:
> Just an idea, as this particular document is always maintained and reviewed by people from the CDI spec; why is it even in the platform?

Oh, I will happily move the responsibility for these parts over to someone else, are you volunteering? :-D

Jokes aside, it is as Ladislav said. The main idea was to avoid having CDI (core) as a responsible party for integration logic of other parts. It really is on the platform level that we should be defining how other parts of EE come together with CDI - not to mention that from implementation perspective is should be either other EE technologies or perhaps integrators (like WFLY) who provide a lot of of the wiring. Be it custom scopes, built-in beans, ...

CDI SE is a little different story - there are no integrators and no other place to put it in. Besides, it is very small API on top of CDI Full.

manovotn (approved):

arjantijms:
> I'm not terribly happy about it, but if we were to move it back to CDI, the only sensible way IMHO would be to create a second spec under the CDI project.

Maybe we should do that. Makes more sense probably. We'll likely have something like that in the Security spec too.

I particularly dislike the platform as a kind of dumping ground for things that don't really belong there, but for which we didn't find a better place yet (maybe because we didn't think hard enough).

For me, the platform is essentially the stuff that an integrator (like GlassFish, WildFly, Liberty, etc) implements under the covers to glue things together and make them work together. 

The conversation scope, for instance, is not like that at all. It's a very visible user facing feature with a specific API. It's still primarily described in the CDI spec, but this text here describes how it applies to Servlet and Faces. Alternatively, maybe Servlet and Faces themselves should both describe this? I'm happy to take over at least the Faces part to the Faces spec.

WDYT?

manovotn:
> Alternatively, maybe Servlet and Faces themselves should both describe this? 

So *this* is exactly what I recall was originally discussed in CDI meetings (years and years ago). We wanted to take those EE parts and move them to their respective specifications. After all, CDI lays down the groundwork for others to do the integration. With that in mind, we, as in CDI spec, didn't want to house all of these parts.
However, that effort winked out when it was clear this would require action from multiple other specs, most of which weren't exactly keen on it.
So we ended up leaving it here :shrug: 

If you are willing to take on the Faces part and place that where it belongs, that's +100 from me

jamezp (approved):

scottmarlow (approved):

arjantijms:
Issue for Servlet spec: https://github.com/jakartaee/servlet/issues/1027

